PR #17, opened 2014-12-26
adds really simple Martini + Facebook example

fredkelly:
Been playing with getting this working using [martini](https://github.com/go-martini/martini). I had to use regular GET params to set the provider (e.g. `?provider=facebook`) so that the builtin `GetProviderName` works. It would be nice to use [`martini.Params`](http://godoc.org/github.com/go-martini/martini#Params) instead but I haven't quite figured out a nice way to override `GetProviderName` and pass that in (still new to Go).

### DIFF
--- a/examples/martini.go
+++ b/examples/martini.go
@@ -1,0 +1,37 @@
+package main
+
+import (
+  "os"
+  "fmt"
+  "net/http"
+
+  "github.com/go-martini/martini"
+  "github.com/markbates/goth"
+  "github.com/markbates/goth/gothic"
+  "github.com/markbates/goth/providers/facebook"
+)
+
+func main() {
+  goth.UseProviders(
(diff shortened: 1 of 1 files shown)


fredkelly:
I got part way there; I saw the issue you mentioned but I'm still a little confused on the conflicting method signatures: I need to pass `martini.Params` but the function expects a request object.

tylerb:
Do something like:

```go

gothic.GetProviderName = func(r *http.Request) (string, error) {
		provider := mux.Vars(r)["provider"]
		if provider == "" {
			return provider, errors.New("missing provider")
		}
		return provider, nil
	}

```

where you are in scope of the martini params. You can use it directly in the closure.

fredkelly:
Gotcha, thanks again!

PR #23, opened 2015-01-28
Switch github provider off the deprecated goauth2 package 

### LINKED ISSUE #20: Support defining which scopes to request for Github provider
Currently, the `authUrl` is hardcoded and there is no mechanism in place to provide which scopes you want to request for the user being authenticated, so by default you only get `public` access.

Would be nice to be able to configure which [Github scopes](https://developer.github.com/v3/oauth/#scopes) are desired.

### LINKED ISSUE #22: Update providers to not use deprecated goauth2 package
The [goauth2](https://code.google.com/p/goauth2/) has been deprecated in favor of the [oauth2](https://godoc.org/golang.org/x/oauth2) package.

We should update the Github, Facebook, and G+ providers accordingly.

cpjolicoeur:
Close #22 
Close #20

### DIFF
--- a/examples/main.go
+++ b/examples/main.go
@@ -10,6 +10,7 @@ import (
 	"github.com/markbates/goth"
 	"github.com/markbates/goth/gothic"
 	"github.com/markbates/goth/providers/facebook"
+	"github.com/markbates/goth/providers/github"
 	"github.com/markbates/goth/providers/gplus"
 	"github.com/markbates/goth/providers/twitter"
 )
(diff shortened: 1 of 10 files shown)


tylerb:
I feel this PR should apply the same updates and changes to all the providers, rather than just one of them. I'd rather not have only one provider that has been updated.

cpjolicoeur:
I can update the others as well.  My current project is only working with Github so that is all that was on my use-case, but I'll try to move the others over as well

tylerb:
That would be much appreciated. Thank you.

markbates:
I agree with Tyler. Thanks Craig.

cpjolicoeur:
The twitter provider was using a different OAuth 1 package and not goauth2 so I didn't do any conversion there.  But the 3 (Facebook, G+, Github) that were using goauth2 have been updated.

cpjolicoeur:
Any more feedback on this?  I think it's ready to go.

tylerb:
Reading through it now.

tylerb:
LGTM

PR #26, opened 2015-02-10
Linkedin provider

nullboundary:
Hi just finished implementing a Linkedin provider. It passes all the tests and I tried it on my own site and it seems to be working. 

Linkedin creates a strange request url filter for fetching user info where you can select very particular fields. I just set it to fetch the fields that will match the goth.User struct. 

Please review the provider and let me know if you have any issues. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ $ go get github.com/markbates/goth
 * Facebook
 * Google+
 * Spotify
+* Linkedin
 
 ## Examples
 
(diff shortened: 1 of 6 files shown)


markbates:
This looks great. Can you just add it to the list of supported providers in the README as well as the example app. Then I'll merge it in. Thanks.

nullboundary:
Ok that should do it. Added Linkedin to both the example and the Readme.

PR #30, opened 2015-03-22
AppEngine compatibility fix

CaptainCodeman:
Passing a hard-coded `oath2.NoContext` to `oauth2.Config.Exchange` breaks AppEngine compatibility following recent changes to golang/appengine and golang/oauth2 to use context.Context (see: golang.org/x/net/context).
Allowing a context to be passed through `gothic.CompleteUserAuth` / `Session.Authorize` allows a context to be supplied, e.g.:

ctx := appengine.NewContext(req)
user, err := gothic.CompleteUserAuth(ctx, res, req)

Even if not running on AppEngine, the context is still useful to allow providers to handle timeouts / cancellation but context.TODO() can be passed for the time being.

### DIFF
--- a/examples/main.go
+++ b/examples/main.go
@@ -16,7 +16,9 @@ import (
 	"github.com/markbates/goth/providers/linkedin"
 	"github.com/markbates/goth/providers/spotify"
 	"github.com/markbates/goth/providers/twitter"
-	)
+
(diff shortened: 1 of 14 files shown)


CaptainCodeman:
I'll look at adding support a different way.

PR #43, opened 2015-10-15
twitter: add authenticate support

oov:
Hi, there!
I've added authentication support to Twitter provider.
Would you review my changes?


markbates:
Can you please add some tests and perhaps an example of how to use this?

oov:
Almost done!
Where shall I add an example?

markbates:
There's an examples folder you can add an example to. Thanks.

oov:
Thank you very much for your advice!
How does that look?

markbates:
Thanks for the PR. It looks good. Thank you.

Issue #33, opened 2015-07-27
Not working correctly for linkedin


idealogyin:
While connecting with linkedin it throws back and error
`You+need+to+pass+the+"state"+parameter`
````
oauth2: cannot fetch token: 400 Bad Request
Response: {"error_description":"missing required parameters, includes an invalid parameter value, parameter more than once. : code","error":"invalid_request"}
````

markbates:
I don't use LinkedIn, so if you open a PR with a fix I'd be happy to merge it in. Thanks.

bentranter:
After some investigation, it looks like this bug is being caused by the redefinition of `gothic.GetState` in `examples/main.go`: https://github.com/markbates/goth/blob/master/examples/main.go#L37-L40. 

LinkedIn seems to be the only provider that makes the `state` parameter required, so omitting the `state` param breaks it. I'll send a PR to fix this at some point, but I'm going to test the other major providers manually first to make sure they all are working in the example before doing so.

Puja-K:
I am still getting this error. Can someone please tell me how to fix it?

PR #45, opened 2015-10-19
Fix LinkedIn auth flow

bentranter:
This changes the way that the state parameter is dealt with. Previously, `GetState` was defined twice: once in `gothic.go`, and again in the example (`examples/main.go`). This redefinition changed the behaviour of `GetState` from returning a string to returning the value of the state parameter from the URL query string. That means that the state parameter wasn't being set by the client during the auth flow. That's okay for most providers, since sending a state parameter is optional... Except for with LinkedIn.

Now, there is a `SetState` and `GetState` function. `SetState` returns a string to be used when setting the `state` parameter as a client during the auth flow, and `GetState` returns the value of the `state` parameter in the query string returned by the provider.

If merged, this will close issue #33. 

I think this PR could use a review from @tylerb, since it looks like he first implemented the `GetState` function and likely knows more about what it's doing/should be doing than I do.

Also, I know that this current behaviour could use a bit more work, since we don't bother to check whether or not the `state` string returned by the provider is the same as the one set by the client, but I figured I should make this PR to fix the LinkedIn flow before working on that. :relaxed:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,8 @@ sudo: false
 go:
   - 1.2
   - 1.3
+  - 1.4
+  - 1.5
   - tip
 
 matrix:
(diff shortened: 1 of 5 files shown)


bentranter:
Closing in favour of #51

PR #48, opened 2015-11-30
update for httprouter

smagic39:
update for github.com/julienschmidt/httprouter


markbates:
Hi @smagic39 thanks for the PR, however, I am not going to accept it. I won't be accepting this for the following reasons:

* The CI tests have failed.
* You committed a `.idea` folder containing files that are only useful for your editor.
* You changed all of the package names from `github.com/markbates/goth` to `github.com/smagic39/goth`.
* I don't think this change is necessary. The current examples show how to use `goth` and `gothic` using a router that is compatible with the standard library. While I think `httprouter` is a great router, it is not standard library compatible so it's not so obvious to those looking at the example code.

Thank you again for the PR.

PR #51, opened 2015-12-11
Fix LinkedIn auth flow, close #33

### LINKED ISSUE #33: Not working correctly for linkedin

bentranter:
This changes the way that the state parameter is dealt with. Previously, `GetState` was defined twice: once in `gothic.go`, and again in the example (`examples/main.go`). This redefinition changed the behaviour of `GetState` from returning a string to returning the value of the state parameter from the URL query string. That means that the state parameter wasn't being set by the client during the auth flow.
That's okay for most providers, since sending a state parameter is optional... Except for with LinkedIn.

Now, there is a `SetState` and `GetState` function. `SetState` returns a string to be used when setting the `state` parameter as a client during the auth flow, and `GetState` returns the value of the `state` parameter in the query string returned by the provider.

(Just a rebase of #45, I didn't realize how badly I botched the rebase on that one).

### DIFF
--- a/examples/main.go
+++ b/examples/main.go
@@ -40,19 +40,12 @@ func main() {
 		bitbucket.New(os.Getenv("BITBUCKET_KEY"), os.Getenv("BITBUCKET_SECRET"), "http://localhost:3000/auth/bitbucket/callback"),
 	)
 
-	// Assign the GetState function variable so we can return the
-	// state string we want to get back at the end of the oauth process.
-	// Only works with facebook and gplus providers.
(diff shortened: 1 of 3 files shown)


Puja-K:
what is the fix? cause i am still getting this error for LinkedIn auth.

PR #52, opened 2015-12-14
Fix LinkedIn state passing

baloo32:
This restores the ability to pass a state parameter to the LinkedIn
provider by sending the state value on the query string. However, if no
state is passed, as LinkedIn required the parameter, a string will be
returned in its place.


bentranter:
Ah this is a million times better than my "fix" haha, I don't know what I was thinking. Awesome work @baloo32!

PR #53, opened 2015-12-19
added the gorilla mux way of reading  out url parameters

Snorlock:
When using a gorilla mux router, this library was not supported in this verison of goth. I added this possibility.

### DIFF
--- a/gothic/gothic.go
+++ b/gothic/gothic.go
@@ -14,6 +14,7 @@ import (
 	"os"
 
 	"github.com/gorilla/sessions"
+	"github.com/gorilla/mux"
 	"github.com/markbates/goth"
 )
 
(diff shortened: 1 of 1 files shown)


markbates:
Thank you for this PR. I'm not going to merge it in, because I would rather have people define their own `GetProviderName` functions, rather than continue to grow the existing one with implementations for every router/mux package out there.

I wish that I could rip out the current implementation and force people to write their own, but unfortunately this would break existing implementations, so I can't do that. :(

Thanks again though.

Snorlock:
Ah yeah I understand now, thx for feedback!

markbates:
If it makes you feel any better the router I use these days isn't compatible with this function, so I have to write my own implementation. :)

tylerb:
I wrote the `GetProviderName` functionality in the first place, as my router wasn't compatible with the way `goth` handles it.

@markbates you certainly could rip out the current way of doing it. It'd be a major feature point bump and need a deprecation notice, but I wouldn't be opposed to it.

Perhaps it would be worthwhile to discuss what has been learned, what we'd like to change and if we'd like to move forward with a 2.0 release.

Snorlock:
@markbates yeah makes me feel alright ;)

Issue #65, opened 2016-03-08
Exotic http lib leak not handled by goth


rgeronimi:
The http lib has an exotic case of connection leak which is not taken care of by goth :
if for some reasons (e.g., ongoing production incident) the target authentication URL is redirecting too many times, the golang http client might return a non-nil response object together with a non-nil error.

This is an under-documented weakness of the Golang version 1 libraries, left for compatibility reasons, causing connection leak in many projects.

Consequently, when checking for an http client error, the goth code should also check if the response is non-nil, and if non-nil close its body.

markbates:
Please submit a PR, I'd happily merge it.

rakesh-eltropy:
@rgeronimi It is not fixed for all the providers. We still have same issue in dropbox, box, digital ocean, lastfm, linkedin, salesforce, spotify, twitch, twitter and yammer providers.

rgeronimi:
Ah thanks, I hacked a simple shell script to search for such errors in my dependencies, and I missed the http Do pattern.
Fixed now, submitting a second PR.

rakesh-eltropy:
@rgeronimi can you please close this.

PR #66, opened 2016-03-08
Fixed http exotic connection leak

rgeronimi:
Fixes issues #65

Although the changes seem simple, I currently don't have the time to configure my environment to test the APIs on the various providers. I will be able to test some of them when my next project starts in 3-4 weeks. Before then you need a test environment to validate the PR.

PS : some linters (errcheck) encourage being explicit with the use or non-use (ignore underscore _) of function return values. E.g., they would force the patch to use this style in the error escape paths :
_ = response.Body.Close()
instead of
response.Body.Close()
I noticed you dont follow the second alternative, so I aligned the patch coding style on yours.


markbates:
Thanks!

PR #68, opened 2016-03-10
Added Refresh Token/Expiry Date of Access Token for applicable providers

rakesh-eltropy:
Hi Mark,
I have Added Refresh Token/Expiry Date of Access Token for applicable providers. There is a change in provider interface where we have additional methods :

- RefreshToken(refreshToken string) (*oauth2.Token, error) //Get new access token based on the refresh token
- RefreshTokenAvailable()(bool) //Refresh token is provided by auth provider or not

I have implemented and tested these for all existing providers.

Kindly review and see if we can merge it to the master.

Thanks,
Rakesh Goyal

### DIFF
--- a/examples/main.go
+++ b/examples/main.go
@@ -112,4 +112,6 @@ var userTemplate = `
 <p>Description: {{.Description}}</p>
 <p>UserID: {{.UserID}}</p>
 <p>AccessToken: {{.AccessToken}}</p>
+<p>ExpiresIn: {{.ExpiresIn}}</p>
+<p>RefreshToken: {{.RefreshToken}}</p>
 `
--- a/examples/main.go
+++ b/examples/main.go
@@ -10,7 +10,9 @@ import (
 	"github.com/gorilla/sessions"
 	"github.com/markbates/goth"
 	"github.com/markbates/goth/gothic"
+	"github.com/markbates/goth/providers/amazon"
 	"github.com/markbates/goth/providers/bitbucket"
+	"github.com/markbates/goth/providers/box"
 	"github.com/markbates/goth/providers/digitalocean"
 	"github.com/markbates/goth/providers/dropbox"
 	"github.com/markbates/goth/providers/facebook"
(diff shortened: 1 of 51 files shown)


bryanl:
@rakesh-eltropy thanks for the PR!

I have a few comments:

1. While `goth` wasn't 100% `golint` compliant before, in most cases, the godoc comments were correct. Could you update these?
2. A few of the files don't look they where formatted by `gofmt` (e.g `providers/digitalocean/digitalocean.go`)
3. Would `ExpiresAt` make more sense than `ExpiresIn`, since you are returning a time and not a duration?

markbates:
Thanks @rakesh-eltropy for this PR! Awesome.

@bryanl thanks for taking the time to review it.

@rakesh-eltropy if you could update based on Bryan's comments that would be great!

zmb3:
Does `RefreshToken()` have to be a method on the `Provider`?

It seems like we could pull that out into a package function so that each provider isn't forced to paste a copy of the same implementation.

Just thinking, if we try to keep providers lean and easy to implement we can expect to get more contributions.

markbates:
@zmb3 I think that's a great idea. What does everyone else think?

bryanl:
I think this is a good idea as well.

markbates:
@rakesh-eltropy can you update to have a package level `RefreshToken()` function?

rakesh-eltropy:
I feel implementation can change for different providers thats why I decided to keep it at provider level.

rakesh-eltropy:
@bryanl thanks for taking the time to review it. I have incorporated your comments and goth is 100 % golint compliant now!

markbates:
Can the implementation actually change for oauth2 refresh tokens between providers? I honestly don't know. Does anybody have any thoughts in this area? My gut says they should all be the same since they're all using the same oath2 package under the covers.

markbates:
An alternative might be to create a `CommonProvider` that can be included into all provider implementations that would provide some default values and functions that could be overwritten on a per provider level if needed.

rakesh-eltropy:
Also Mark, we would need tokenURL to make refresh token function work. Its not a great idea to ask user to pass tokenURL as it will be different for each provider and we want to keep it seamless for the library user.

markbates:
@rakesh-eltropy i would assume you would take the provider as an argument to the `RefreshToken` function, which would solve that problem.

The biggest problem I see is that `RefreshToken` uses `p.config`, which is private, which would make a package level function not doable without adding a `Config()` function to the `Provider` implementation.

markbates:
It doesn't sound like having a package level `RefreshToken` function makes sense, after thinking about it.

If no one has any other objections I'm going to merge this PR. Anyone?

bryanl:
@markbates merge away

bketelsen:
wow!

rakesh-eltropy:
Thanks a ton Mark!

markbates:
@rakesh-eltropy thank you!! This was a great PR!

PR #70, opened 2016-03-13
added auto state generation

n7fury:


n7fury:
ops sorry i have not completed testing, i'm very new to the whole github stuff

n7fury:
this pull request should automatically generate a random 16 byte state string and automatically add it if the user is using the Gothic package? The only issue i can foresee is if given an incorrect state it will not show the correct information but will not display an error I'm unsure how to fix this.

Ps. Sorry for muck up this is my first pull request

rakesh-eltropy:
@n7fury I think, we should change SetState function to return auto generated state string if its not available in request. What do you think @markbates ?

n7fury:
@rakesh-eltropy @markbates anything else i can do, i :heart_eyes: this project and would love to help more.

rakesh-eltropy:
@n7fury Thanks for the PR! Sure, we would love to have more contributions. Please see if you can work on following providers:

- Coursera
- Shutter Stock
- Egnyte
- Pcloud
- IBM Connections

I am currently working on more providers and idea is to have as many providers as possible. We want goth to be goto library for any auth in Golang.

PR #73, opened 2016-03-26
Use login name as nickname for GitHub provider

### LINKED ISSUE #72: Github provider does not include username in User object
The github user returned from FetchUser does not include the username anywhere. It is in the api response as 'login'.

Should the `NickName` field be set to that rather than the person's full name?

bentranter:
Would close #72 if merged

The `nickname` field for GitHub previously used the user's name. It was suggested that the nickname field could be used to store the user's login instead, which seems like a good idea. Tested this quickly on my local machine with the GitHub API as well to make sure this didn't break anything.


rushairer:
Maybe   login -> Name, name -> NickName is better than it.

Issue #78, opened 2016-04-21
gothic does not throw any errors, if the callback does not contain a `code` value


geofffranks:
I have a mis-configured github provider being used from gothic. If I misconfigure my callback, so that it does not match what is configured in github, the requests to my callback handler look like this:

```
GET /v1/auth/github/callback?error=redirect_uri_mismatch&error_description=The+redirect_uri+MUST+match+the+registered+callback+URL+for+this+application.&error_uri=https%3A%2F%2Fdeveloper.github.com%2Fv3%2Foauth%2F%23redirect-uri-mismatch&state=41790af3-2ac4-4223-96e6-6e8d38621b1e HTTP/1.1
```

When I issue `gothic.CompleteUserAuth()`, it returns no errors, and a user object of 
`{map[message:Bad credentials documentation_url:https://developer.github.com/v3] github     0      0001-01-01 00:00:00 +0000 UTC}`

I feel like situations like this should cause a failure, since otherwise, it means anyone who hits the callback with the correct state parameter is be authenticated, whether the provider thinks they should have been or not.

Where should this get fixed? The provider code for github (and possibly others)? gothic?  Is this an underlying issue with oauth2.Extract()?

In the mean time, I've worked around this by requiring the 'code' param to be set in the callback before passing on to gothic.

geofffranks:
Is this a problem with github returning a 200 with a payload error that isn't being inspected?

markbates:
Most likely that is what the problem is. A PR would be very appreciated. :)

PR #79, opened 2016-04-21
Validate tokens from provider during the Authorize() call

geofffranks:
Addresses #78 by updating all providers using the oauth2 framework to validate the tokens they receive from the provider, and throw an error if the token is invalid.


markbates:
Awesome! Thanks!

PR #83, opened 2016-05-17
Find provider in mux vars

plutov:
I am using mux router, and Query().Get("provider") is always empty. So I added a check in mux Vars.

### DIFF
--- a/gothic/gothic.go
+++ b/gothic/gothic.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"os"
 
+	"github.com/gorilla/mux"
 	"github.com/gorilla/sessions"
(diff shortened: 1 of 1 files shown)


rbrumby:
I'd vote against adding this (sorry plutov ;)). I found the same issue myself but the original author already allows us to provide an override function by:
gothic.GetProviderName = func(r *http.Request) (name string, err error) {
		vars := mux.Vars(r)
		name = vars["provider"]
		return
	}
Including the code in this pull request creates a dependency on gorilla mux which creates bloat. With numerous routers out there we wouldn't want to end up dependent on them all.

plutov:
@rbrumby , thanks for this note that we can override this func.

PR #84, opened 2016-06-09
Support GitHub Enterprise

jcbwlkr:
One of my clients wants to add GitHub authentication to their site but they use a hosted version which is not at github.com. The GitHub provider in this package uses hard coded URLs for authentication, tokens, and the API and they are not exposed in a way I can change them from another package.

I created a func `NewEnterprise` which wraps around `New` but changes the urls before returning. I also added a field to `Provider` of the URL to use when fetching the user profile.

I couldn't see a reasonable way of testing that `NewEnterprise` sets the token url appropriately so I didn't. If you can think of any changes you would like I'd be happy to make them.

### DIFF
--- a/providers/github/github.go
+++ b/providers/github/github.go
@@ -15,10 +15,17 @@ import (
 	"strconv"
 )
 
-const (
-	authURL         string = "https://github.com/login/oauth/authorize"
-	tokenURL        string = "https://github.com/login/oauth/access_token"
-	endpointProfile string = "https://api.github.com/user"
+// These vars define the Authentication, Token, and API URLS for GitHub. If
+// using GitHub enterprise you should change these values before calling New.
+//
+// Examples:
+//	github.AuthURL = "https://github.acme.com/login/oauth/authorize
+//	github.TokenURL = "https://github.acme.com/login/oauth/access_token
+//	github.ProfileURL = "https://github.acme.com/api/v3/user
+var (
+	AuthURL    = "https://github.com/login/oauth/authorize"
+	TokenURL   = "https://github.com/login/oauth/access_token"
+	ProfileURL = "https://api.github.com/user"
 )
 
 // New creates a new Github provider, and sets up important connection details.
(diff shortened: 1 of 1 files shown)


markbates:
I think the easier solution would be change the const urls to var and make them public. Would that also solve your problem?

jcbwlkr:
haha umm yeah I guess I shot right past that idea. That is definitely much simpler and yeah I can work with that. The only problem I can see with using package vars is it would mean you couldn't simultaneously support logging in against multiple instances but that seems pretty unlikely.

I can go ahead and change this PR to just expose those vars if you would like.

markbates:
Yeah, let's start with the simplest change first. I thought of the issue of wanting to use both GitHub and github enterprise, but we can cross that bridge when we come to it. :)

jcbwlkr:
That is done. Before you merge do you want to change the names of any of those vars? Once they're exported they're part of the public API and changing them later might break somebody. Specifically I'm thinking of the var `EndpointProfile` just following a different pattern from the other two (thought it may be consistent with the other packages, I haven't looked)

providers/github/github.go
```diff
@@ -15,10 +15,17 @@ import (
...
+// using GitHub enterprise you should change these values before calling New.
+//
+// Examples:
+//	github.AuthURL = "https://github.acme.com/login/oauth/authorize
+//	github.TokenURL = "https://github.acme.com/login/oauth/access_token
+//	github.EndpointProfile = "https://github.acme.com/api/v3/user
+var (
+	AuthURL         = "https://github.com/login/oauth/authorize"
+	TokenURL        = "https://github.com/login/oauth/access_token"
+	EndpointProfile = "https://api.github.com/user"
```
markbates: Change to `ProfileURL`.

jcbwlkr: Done

markbates:
Thanks!

jcbwlkr:
Awesome! Thanks for this package and I'm glad to have helped :)

markbates:
Thank you. It's always great when others help make something you wrote better!

PR #87, opened 2016-06-24
Replace direct usage of http.DefaultClient in providers and oauth2 calls

felixLam:
Some providers already used a custom httpClient. This brings the `Client` field to all providers and uses a fallback method to provide the `http.DefaultClient` if no client is set. 

This also uses the custom httpClient for the oauth2 calls via the context.

The change should allow better inversion of control during unit tests by providing a custom client.

I also ran go format on some files to clean up the import statements.

### DIFF
--- a/provider.go
+++ b/provider.go
@@ -1,7 +1,12 @@
 package goth
 
-import "fmt"
-import "golang.org/x/oauth2"
+import (
+	"fmt"
+	"net/http"
+
+	"golang.org/x/net/context"
+	"golang.org/x/oauth2"
+)
 
 // Provider needs to be implemented for each 3rd party authentication provider
 // e.g. Facebook, Twitter, etc...
@@ -46,4 +51,20 @@ func GetProvider(name string) (Provider, error) {
 // This is useful, mostly, for testing purposes.
 func ClearProviders() {
 	providers = Providers{}
+}
+
+// ContextForClient provides a context for use with oauth2.
+func ContextForClient(h *http.Client) context.Context {
+	if h == nil {
+		return oauth2.NoContext
+	}
+	return context.WithValue(oauth2.NoContext, oauth2.HTTPClient, h)
+}
+
+// HTTPClientWithFallBack to be used in all fetch operations.
+func HTTPClientWithFallBack(h *http.Client) *http.Client {
+	if h != nil {
+		return h
+	}
+	return http.DefaultClient
 }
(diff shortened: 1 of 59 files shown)


felixLam:
This does not yet cascade down to the oauth2 level. For that I would need to change the context of the provider's oauth configuration.

markbates:
Why not have the `Provider` interface have a `Client() *http.Client` function as part of the definition. I feel like that would be the better abstraction.

felixLam:
I thought so too, but you would still need a field for each provider in order yo let you set it externally. The Client field was also already available in some providers. The current implementation requires less code duplication unless I am missing something.

felixLam:
@markbates Did you have time to look at this PR again? If you have any suggestions I am more than happy to implement them.

markbates:
I'm sorry, I haven't. I'm traveling through Italy right now with my family, so I don't have the time nor the Internet to really investigate. I plan on checking this out when I get to gopher con the end of next week.

felixLam:
Enjoy your holidays 🇮🇹 and make sure to watch the match with the locals tomorrow ⚽

felixLam:
Hi @markbates  any news on this PR?

felixLam:
@markbates any news on this PR? Is it worthwhile updating it to resolve conflicts?

willemvd:
@markbates This would really help us out in [Gitea](https://github.com/go-gitea/gitea) to set custom fields in the Client!
Do you have any plans (and when) to implement these?

felixLam:
@willemvd it is implemented on our branch in this PR. If you care and plan on using it I can rebase it on the current master. Since @markbates has gone radio silent here, I haven't looked into it.

willemvd:
@felixLam since @markbates is actually doing stuff on the master (last time 2 days ago) I hope he will respond on this, but a rebase is always welcome!

felixLam:
@willemvd I merged master into this PR (didn't do a rebase as we have our own branch on top of this PR). Let me know if this works.

willemvd:
@felixLam found out that the new introduced auth0 provider is not updated with the new Provider.Client field

felixLam:
Weird that this is not raised by the CI. I ran the test locally and they passed, can you let me know what command you used that failed?

felixLam:
Simply go build ?

willemvd:
I just looked at the code and saw that it was missing (think also in the session.go) 😄 
https://github.com/iosphere/goth/blob/feature/httpClient/providers/auth0/auth0.go#L23-L27

willemvd:
@markbates Thanks!

PR #95, opened 2016-08-21
Fix CloudFoundry panic due to uninitialized http.Client

rbrumby:
Also removed a fmt.Printf form cloudfoundry/session.go


markbates:
Thanks!

PR #102, opened 2016-10-21
Rename bio to about

zusyed:
In the v2.8 update of Facebook, the bio field has been replaced with
about.

### DIFF
--- a/providers/facebook/facebook.go
+++ b/providers/facebook/facebook.go
@@ -6,18 +6,19 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
-	"github.com/markbates/goth"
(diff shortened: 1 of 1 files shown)


markbates:
Will this break for people using an older version of the API? If so, then the patch should support both `bio` and `about`.

zusyed:
The Facebook API that is called to fetch a user is 
https://graph.facebook.com/me?fields=email,first_name,last_name,link,about,id,name,picture,location which does not specify any versions. So I assume it will always call the latest version of Facebook APIs. I do not see how consumers of this Facebook provider would use an older version of the API.

markbates:
Thanks!

PR #103, opened 2016-10-24
HTTPS Example

chandler767:
This example is the same as the existing example but it uses HTTPS with HTTP Strict Transport Security.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ generated/
 *.swp
 .vscode/launch.json
 .vscode/settings.json
+*.crt
+*.csr
+*.key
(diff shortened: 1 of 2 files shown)


markbates:
The tests are failing. Also, I don't understand the point of this example. What does running the local server in https provide in terms of insight using goth?

chandler767:
It doesn't, It just gives you an example that puts you on the right track to keep user info secure. Since you're handling user information and probably going to implement goth with https anyway, why shouldn't the example use https?

The http example is fast and simple to try, but I think an https example gives you an idea of what you should be doing when you go to build your application with goth. If you'd prefer to only have the http example just because it's a better example to highlight goth, I understand. My thinking is that I'd want an example that is similar to what I want to be doing in production.

It looks like the test is failing because it's not including "crypto/tls" needed for my example.

PR #106, opened 2016-11-07
Update LinkedIn OAuth Endpoints to latest correct version

baloo32:
Current implementation is using outdated LinkedIn endpoints (as per https://developer.linkedin.com/docs/oauth2) which return URL Fragments on authorization. Updated to latest endpoints.


markbates:
Thanks!

PR #107, opened 2016-12-07
Handle non-200 responses from GitHub API endpoints

jcbwlkr:
This shouldn't matter if you're using regular GitHub but if you change the URLs to point to your Enterprise instance and you mess it up it would previously fail silently and you'd be logged in with a zero value User.

I added a check that the response is `http.StatusOK`. I tested this by forcing the `github.ProfileURL` to something I know will 404 and it now errors out. I did the same for the newly added email check.

I also moved a `defer` after an `err` check and updated some documentation.

### DIFF
--- a/providers/github/github.go
+++ b/providers/github/github.go
@@ -6,6 +6,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


markbates:
Great! Thanks!

PR #113, opened 2016-12-29
proposal: app engine support

jtolds:
App Engine unfortunately can't make outbound requests without using
the appengine urlfetch library and a request-specific context.

I don't see a way around this other than to change the signature of
the Provider and Session interfaces.

This really has a huge amount of downsides obviously, but before I
go and fix stuff for providers besides Facebook, G+, and Twitter
(Twitter support requires https://github.com/mrjones/oauth/pull/60
to get merged), I figured I'd open a pull request and discuss the
tradeoffs.

Unfortunately I think we need to do something like this to support
App Engine. Perhaps this is an opportunity to version goth using
gopkg.in or something?


jtolds:
Yeah, there's no way the builder passes. Let me know if I should spend time fixing all the other providers.

jtolds:
Of course, the other option is to add methods to all the interfaces. This makes provider implementation harder, but maybe that's worth the backwards compatibility.

tylerb:
Perhaps some kind of global `Requester` interface/object that defaults to what is in there now, but can be overridden by the user-programmer?

jtolds:
@tylerb: that's included in the above patchset. the problem is that the overrideable global requester requires knowing request specific context information. on app engine, you can't make an outbound request without tying it to the request that kicked off your app engine code

jtolds:
I'll refactor to not break backwards compat.

jtolds:
Okay, new commit, still only touched Facebook, G+, and Twitter, but should be backwards compatible with downstream libraries.

jtolds:
If there's a chance this approach might be accepted, I'll get to work fixing all the other providers.

asawilliams:
Is there a work around for getting goth to work with GAE or should we wait until this gets merged in?

PR #118, opened 2017-01-09
Add the link to godoc in the header title

chilts:


chilts:
Feel free to ignore this. Or we can leave it in the header and I can put the Docs header and instructions back into the main body of the text. Your call? :)

markbates:
I prefer it in the title like you've added. Thanks!

PR #119, opened 2017-01-09
Document that UseProviders can be called multiple times

chilts:


chilts:
Had to figure this out since I'm calling it with a provider, but only if that provider has been configured in the environment (for my use-case). Thinking it might be good to document.

chilts:
Forgot to note that I also changed `sets` to `adds` since the old list isn't overwritten. Just noting it here since it doesn't show up as well as the additional diffs and wanted to point it out in case you didn't like it.

chilts:
Lol, you merged before I adds that comment. :) Thanks.

PR #125, opened 2017-01-20
add email_verified check for intercom

mmartinic:
Adds check if user email is verified and fail with "User email not verified" error if not.
Also adds custom user agent

### DIFF
--- a/providers/intercom/intercom.go
+++ b/providers/intercom/intercom.go
@@ -120,10 +120,6 @@ func userFromReader(reader io.Reader, user *goth.User) error {
 		return err
 	}
 
-	if u.EmailVerified != true {
-		return errors.New("User email not verified")
-	}
-
 	user.Name = u.Name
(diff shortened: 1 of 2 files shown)


providers/intercom/intercom.go
```diff
@@ -118,6 +120,10 @@ func userFromReader(reader io.Reader, user *goth.User) error {
 		return err
 	}
 
+	if u.EmailVerified != true {
+		return errors.New("User email not verified")
```
markbates: By returning an error here, you're breaking existing functionality. Plus, it would be better to let the apps decide to do with a user that is unverified than to an error.

mmartinic: I updated the PR.
Apps can use `user.RawData["email_verified"]` to get that value?

markbates: Yeah, it'll be in the raw data so they can get access to it.

PR #126, opened 2017-01-23
Adding the ability to inject own GetProviderName func

kjda:


markbates:
I'm sure I understand this PR. You could already change the implemention of GetProviderName by assigning a new function to it. Can you please explain? Thanks.

kjda:
I was just thinking this is a more explicit way of offering this ability, mutating package exports directly didn't seem like the best practice to me at first sight

willemvd:
Sorry wrong interpreted this issue, no relation to #130 
Btw , had to do this myself as well because of macaron routing engine.
Used this approach : 
func init() {
	gothic.GetProviderName = func(req *http.Request) (string, error) {
		return req.Header.Get(providerHeaderKey), nil
	}
}
...
func Auth(provider string, request *http.Request, response http.ResponseWriter) {
	request.Header.Set(providerHeaderKey, provider)
	gothic.BeginAuthHandler(response, request)
}

markbates:
I would prefer to see a documentation PR for this, rather than a code PR. Thanks, but I'm going to close this.

PR #129, opened 2017-01-26
Mandatory user

### LINKED ISSUE #128: UserID is not always filled
Make UserID a mandatory field so applications using goth can trust on the fact that this is always filled (and can use it as a mandatory field)

Will create a PR for it

willemvd:
add UserID to providers that haven't implemented it 
- fitbit
- instagram
- faux (to be complete 😄 )

Should fix #128


markbates:
Cool. Thanks.

Issue #123, opened 2017-01-18
Use new Client() in all cases


willemvd:
Seems that not all http.Get (and maybe .Post?) are replaced with the new Client() from the provider.
Example: https://github.com/markbates/goth/blob/master/providers/github/github.go#L154

markbates:
I replaced all the ones that had been set up by that last PR Felix set up, so that must've missed a bunch of places. 

PR? Please?

willemvd:
Ok, will try to do that tomorrow but not sure if I will get to it :)

PR #130, opened 2017-01-26
Support multiple instances of one provider type

willemvd:
To be able to register the same provider twice we need to be able to override the provider name.
Don't want this to be a new func registered to the provider.Name but be inline with Client ( #123 ) 
Downside is that existing calls to provider.Name() will break (easy to replace with GetName() )

This will maybe also fix issue #126 ?

### DIFF
--- a/provider.go
+++ b/provider.go
@@ -11,7 +11,7 @@ import (
 // Provider needs to be implemented for each 3rd party authentication provider
 // e.g. Facebook, Twitter, etc...
 type Provider interface {
-	Name() string
+	GetName() string
 	BeginAuth(state string) (Session, error)
 	UnmarshalSession(string) (Session, error)
 	FetchUser(Session) (User, error)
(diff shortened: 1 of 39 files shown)


willemvd:
No influence on #126

markbates:
I'm not ok with a breaking API change. There needs to be a different way of doing this. Perhaps a setter? Or people could just wrap the provider in a type that reimplements the name function.

willemvd:
Think you end up with something like this
```
		provider := github.New(clientID, clientSecret, "http://localhost:3000/auth/github/callback")
		provider.Name = func() string {
			return "github" // or any other name
		}
```
but when you have multiple types of providers this might get messy with lots of extra func definitions.
That is why I have suggested this approach, but keeping Name() with some extra SetName() might do the trick

willemvd:
I will update the PR with that approach

markbates:
So my problem now is none of the providers have a default name. We need to return a default name for each provider unless one is set. If this PR was merged it would break existing apps, unless I'm missing something.

willemvd:
All the New methods set the default name in the Provider 
example: https://github.com/willemvd/goth/blob/name-method/providers/amazon/amazon.go#L41

willemvd:
Btw see a new conflict on intercom, will merge that

markbates:
Ah! Awesome! I missed that. Sorry.

willemvd:
NP 😄

Issue #132, opened 2017-01-31
Get goth.User without re-authenticating every time


willemvd:
With the current set of functions in the gothic.* we don't have an option to retrieve the user without going through the whole authentication flow.

Current setup to get user details at login:
1. start login flow with BeginAuthHandler or GetAuthURL (if you need to do custom error handling)
2. handle callback and get the goth.User with CompleteUserAuth

Step 1 will always try to completely authenticate the user again even if he use got a session with a valid accessToken and only did a logout

To prevent this, the calling application should first do a CompleteUserAuth request and see if the goth.User can be fetched from there. To support this we need to store the AccessToken in the session and add a extra FetchUser call in the CompleteUserAuth before doing sess.Authorize

What do you think of this?
Will create a PR to show the suggested fix (without introducing all kind of extra stuff in each provider to prevent this re-authenticating)

PR #134, opened 2017-02-01
Get user without reauthenticating

willemvd:
PR for issue #132 
Updated the example with the way users can first check if the user is already authenticated and has a valid session

To make it work with all providers I have added checks in the FetchUser methods to check if the accessToken is set and if set, if retrieval of the provider User was successful

The Yammer provider is updated to use the Yammer current user API to get the user information. This way it can work like the other OAuth providers (otherwise to determine if the session was still valid would become to difficult)

Additional fix inside this PR is the example of OpenID Connect which seems to be broken when no settings where given

### DIFF
--- a/examples/main.go
+++ b/examples/main.go
@@ -97,10 +97,15 @@ func main() {
 
 		//Auth0 allocates domain per customer, a domain must be provided for auth0 to work
 		auth0.New(os.Getenv("AUTH0_KEY"), os.Getenv("AUTH0_SECRET"), "http://localhost:3000/auth/auth0/callback", os.Getenv("AUTH0_DOMAIN")),
-
-		//OpenID Connect is based on OpenID Connect Auto Discovery URL (https://openid.net/specs/openid-connect-discovery-1_0-17.html)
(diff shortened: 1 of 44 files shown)


willemvd:
Btw I also changed the way the sessions are stored.
Each provider will now have it's own session, because when using 1 session it would be impossible to see if the logged in session was belonging to the right provider

markbates:
So, if I'm reading this correctly what is happening here is if a the session has an AccessToken in it we skip going back to the provider and just pretend that we do. If there isn't an AccessToken we go to the provider as normal. Is this correct?

willemvd:
Almost 😄  
In the CompleteUserAuth we add a new FetchUser method call in which we check if AccessToken exists and the call to the profile service of the provider is successful and return that result.
If one of both fail, we continue the normal flow

The only point is that users should change there flow to first try to do the CompleteUserAuth and if that fails , start the BeginAuthHandler or GetAuthURL

willemvd:
@markbates could you have a look at this? Would like to get this merged

willemvd:
Just tried to add new providers to my local test setup and apparently it failed on new sessions.
Added the ignoring of useless errors, since empty sessions will be created when they are non existing

markbates:
👍

Issue #133, opened 2017-02-01
securecookie: the value is too long


willemvd:
In the OpenID Connect provider the id_token is stored in the session because this contains the user information. Storing the session can cause the error "securecookie: the value is too long" when there is a lot of data stored in the id_token field

willemvd:
We should think of a way to store more information if it doesn't fit in the cookie.
Maybe introduce a new Store which uses disk storage combined with a private key to secure the data?
@markbates Any suggestions?

markbates:
I wouldn't want to change the default store for a cookie store, as that is easy for most people to get up and running with, plus, this only appears to be a problem with the OpenID provider, as far as I can tell.

Since we're using http://www.gorillatoolkit.org/pkg/sessions it should be easy enough for people to plugin their own session stores if the cookie store isn't working for them.

RubenGarcia:
I am also affected.

PR #141, opened 2017-02-20
example update to prevent "securecookie: the value is too long"

willemvd:
I have updated the example to show how to prevent "securecookie: the value is too long" exceptions


willemvd:
No need to create a custom Store for #133 since this will do

egbertp:
@willemvd Thanks for raising the issue and for sharing your working example with us! 

I've implemented the Azure AD as the provider and I needed to extend the session store to get it working. You can find my working `auth.go` file [in this gist](https://gist.github.com/egbertp/682350fa9f1ea42705fbc30f32152e99)

PR #145, opened 2017-02-23
Provide logout convenience method

neumayer:
* Extend faux provider to keep provider name
* Invalidate session cookie and session value on logout
* Add tests

### DIFF
--- a/gothic/gothic.go
+++ b/gothic/gothic.go
@@ -184,27 +184,7 @@ func Logout(res http.ResponseWriter, req *http.Request) error {
 		return err
 	}
 
-	provider, err := goth.GetProvider(providerName)
-	if err != nil {
-		return err
-	}
-
-	value, err := getFromSession(providerName, req)
-	if err != nil {
-		return err
-	}
-
-	sess, err := provider.UnmarshalSession(value)
-	if err != nil {
-		return err
-	}
-
-	user, err := provider.FetchUser(sess)
-	if err != nil {
-		return err
-	}
-
-	session, err := Store.Get(req, user.Provider+SessionName)
+	session, err := Store.Get(req, providerName+SessionName)
 	if err != nil {
 		return err
 	}
--- a/gothic/gothic.go
+++ b/gothic/gothic.go
@@ -176,6 +176,47 @@ var CompleteUserAuth = func(res http.ResponseWriter, req *http.Request) (goth.Us
 	return provider.FetchUser(sess)
 }
 
+// Logout invalidates a user session.
+func Logout(res http.ResponseWriter, req *http.Request) error {
+
+	providerName, err := GetProviderName(req)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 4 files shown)


gothic/gothic.go
```diff
@@ -176,6 +176,51 @@ var CompleteUserAuth = func(res http.ResponseWriter, req *http.Request) (goth.Us
...
+	if err != nil {
+		return err
+	}
+
+	user, err := provider.FetchUser(sess)
+	if err != nil {
+		return err
+	}
+
+	if err != nil {
```
willemvd: why is there a duplicate if err != nil ?

gothic/gothic.go
```diff
@@ -176,6 +176,51 @@ var CompleteUserAuth = func(res http.ResponseWriter, req *http.Request) (goth.Us
...
+		return err
+	}
+
+	ss, err := Store.Get(req, user.Provider+SessionName)
```
willemvd: Please change "ss" to a more readable "session"

gothic/gothic.go
neumayer: Done.

gothic/gothic.go
neumayer: Done.

willemvd:
While testing the PR I found that the logout code is also doing a `FetchUser.`
That seems to be useless since it can only give errors in case something went wrong (user already removed the attached OAuth2 application in his provider profile for example) and logout will then fail.

The `session, err := Store.Get(req, user.Provider+SessionName)`  -> https://github.com/neumayer/goth/blob/9ccab3039c05b868581021906f42cf86788185de/gothic/gothic.go#L207 call should be done with the providerName so there is no need to get the user it self.
This basically removes all code between the `GetProviderName` error check and the `Store.Get`

willemvd:
Btw, not sure what happened last time, but your commits to update the PR are gone. Please only commit and push the new code and not force push to override stuff. Otherwise tracking the changes is difficult

providers/faux/faux.go
```diff
@@ -43,9 +44,10 @@ func (p *Provider) BeginAuth(state string) (goth.Session, error) {
 func (p *Provider) FetchUser(session goth.Session) (goth.User, error) {
 	sess := session.(*Session)
 	return goth.User{
-		UserID:          sess.ID,
-		Name:            sess.Name,
-		Email:           sess.Email,
+		UserID:   sess.ID,
+		Name:     sess.Name,
+		Email:    sess.Email,
+		Provider: sess.Provider,
```
neumayer: This line is not strictly needed any longer with the simplified method calls. I still think it is better to keep it to be consistent in the test provider.

willemvd: Not sure what you mean here?
There is no need to store the Provider inside of the session, since retrieving the session is already based on the provider name and so you will always know that.

So think it is better to remove it

neumayer: I completely removed it from the faux provider implementation.

PR #150, opened 2017-03-29
Add functionality to revoke tokens

neumayer:
* Add revoke method to provider interface
* Implement method for github and gplus
* Add tests
* Add noop implementation to other providers

### DIFF
--- a/examples/main.go
+++ b/examples/main.go
@@ -179,7 +179,11 @@ func main() {
 	})
 
 	p.Get("/logout/{provider}", func(res http.ResponseWriter, req *http.Request) {
-		gothic.Logout(res, req)
+		err := gothic.Logout(res, req)
+		if err != nil {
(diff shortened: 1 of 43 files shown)


neumayer:
Anyone interested?

markbates:
@neumayer my apologies for not looking at this sooner. I think this is fantastic, and definitely wanted. Please continue with the implementations, we need implementations for all the providers that support it before it can be merged. Thanks.

neumayer:
Hehe, that includes me getting accounts with up to 32 identity providers, so it might take a couple of days :-)

But I'll try to keep working on it when I find the time.

neumayer:
I looked at both twitch and amazon today.

While I hunted down some secret amazon api endpoint, there's no documentation for it and I didn't manage to authenticate to it. Since it's not documented at all I think it's not worth pursuing that. I added a short comment about this.

PR #151, opened 2017-04-04
Add xero

TheRegan:
Added support for Xero as a provider.


markbates:
Thanks! Sorry for the delay in merging this.

PR #154, opened 2017-05-03
example how to set env vars

xenu256:


markbates:
You shouldn't be setting environment variables like that. They should be externally through the environment. You don't want to commit that information in your source code because with that information people can gain access to your oauth applications. I can't accept this pr as it suggests a bad practice. I will accept a pr that adds documentation around setting them correctly. Thanks.

PR #159, opened 2017-06-12
Added state token checks to prevent CSRF

ncantelmo:
This pull request adds a state token check to prevent CSRF attacks.

The check runs during CompleteUserAuth(), immediately after the original session has been unmarshalled. The original state token can be pulled from goth.Session.GetAuthURL()'s query params, so no changes to individual providers appears to be needed.

I've added a test that covers both matching and mismatching state tokens. To get the test to run, I needed to added a bit more detail to the faux provider, change the test Store to use the dummy ProviderStore (so I could simulate setting cookies on incoming requests), and disable parallel execution of tests. The tests still run on my system in 0.012s, so I'm hoping that's not an issue.

### DIFF
--- a/gothic/gothic.go
+++ b/gothic/gothic.go
@@ -180,9 +180,8 @@ var CompleteUserAuth = func(res http.ResponseWriter, req *http.Request) (goth.Us
 	return provider.FetchUser(sess)
 }
 
-// validateState checks whether a state token was stored
-// with the original session and, if so, makes sure that
-// the current request contains a matching state value.
+// validateState ensures that the state token param from the original
+// AuthURL matches the one included in the current (callback) request.
 func validateState(req *http.Request, sess goth.Session) error {
 	rawAuthURL, err := sess.GetAuthURL()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


markbates:
Excellent! Thanks!

im-kulikov:
@ncantelmo @markbates this PR breaks Twitter Provider...

im-kulikov:
https://github.com/markbates/goth/blob/master/providers/twitter/twitter.go#L87
```go
// BeginAuth asks Twitter for an authentication end-point and a request token for a session.
// Twitter does not support the "state" variable.
func (p *Provider) BeginAuth(state string) (goth.Session, error) {
	requestToken, url, err := p.consumer.GetRequestTokenAndUrl(p.CallbackURL)
	session := &Session{
		AuthURL:      url,
		RequestToken: requestToken,
	}
	return session, err
}
```

im-kulikov:
mb this can fix a problem (for me it's work):

```go
func (p *Provider) BeginAuth(state string) (goth.Session, error) {
	requestToken, url, err := p.consumer.GetRequestTokenAndUrl(p.CallbackURL)
	session := &Session{
		AuthURL:      url + "&state=" + state,
		RequestToken: requestToken,
	}
	return session, err
}
```

markbates:
Can you submit a PR with the fix please? Thanks.

im-kulikov:
@markbates if it's ok - see https://github.com/im-kulikov/goth/commit/9f6e3ff01f2853d909dfcaea2c57bf04592e3e87

Issue #160, opened 2017-06-13
Stronger default cookie security settings for gothic.Store


ncantelmo:
By default, gothic's CookieStore is created without additional configuration beyond what's set by `sessions.NewCookieStore([]byte(key))`.

That causes gothic to produce cookies with options:
```go
&Options{
   Path:   "/",
   Domain: "",
   MaxAge: 86400 * 30,
   Secure: false,
   HttpOnly: false,
 }
```

The biggest issue here is the `HttpOnly` flag being false. Enabling that by default would be easy, shouldn't break anything, and would help prevent a large class of XSS attacks.

The other item that's worth noting is the `Secure` flag, which requires cookies to be sent over https. Most modern sites served over TLS will want that flag set in production.

Both of these flags can be enabled using gothic currently, if `gothic.Store` is overridden:

```go
isProd := false
store := sessions.NewCookieStore(key)
store.MaxAge(86400 * 30)  // 30 days
store.Options.Path = "/"
store.Options.HttpOnly = true
if isProd {
  store.Options.Secure = true
}
gothic.Store = store
```

Because this can be configured, I don't think any API changes need to be made to gothic. At the same time, since a lot of gothic users will stick with the defaults, I believe the `HttpOnly` flag should be set by default. It would probably also be a good idea to update the example/README to show how to configure the other options, in particular the `Secure` flag.

markbates:
Sounds good. Can you do a PR for the `HttpOnly` flag and an update to the README?

ace03uec:
This has been merged. This issue should be closed.

PR #161, opened 2017-06-13
Set HttpOnly on cookies from gothic.Store by default

ncantelmo:
Addresses #160


markbates:
Love it! Thanks for your help on these security issues.

PR #172, opened 2017-08-22
Added support to get provider name from request URL's go context

KunalSarkar:
The function getProviderName currently support's fetching "provider' parameter only from gorilla's context. Added support to fetch the same from Go's context.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,8 @@ language: go
 sudo: false
 
 go:
-  - 1.5
-  - 1.6
   - 1.7
+  - 1.8
   - tip
 
 matrix:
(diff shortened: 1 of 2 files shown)


KunalSarkar:
Thanks for pointing that out Mark. My bad. I have corrected it now. :)

markbates:
The tests are failing on Travis because `context` was introduced in `1.7` so the tests for `1.5` and `1.6` fail. Can you upgrade the `.travis.yml` file to support `1.7`, `1.8`, and `tip`? Thanks.

KunalSarkar:
Sure, Mark. Done 👍

markbates:
Brilliant! Thanks!

PR #175, opened 2017-09-10
Gothic nonce state

andyhaskell:
This PR updates the state generation in Gothic's `BeginAuth` so that if a request has no `state` query parameter, before the redirect to the provider's authentication URL, a nonce state is generated and added to the query parameters, rather than the string literal "state". This is so that the state parameter is unguessable to protect against CSRF attacks, as advised in https://auth0.com/docs/protocols/oauth2/oauth-state#keep-reading.

Looking at the API documentations for providers, most providers do support the state parameter, but there were a few where I found no explicit mention in the docs of a state parameter. However, providers like Xero and Twitter are already documented to not support the state parameter, and their auth URLs in Goth therefore do not include that param, so to the best of my knowledge, this PR should not be a breaking change.

### DIFF
--- a/gothic/gothic.go
+++ b/gothic/gothic.go
@@ -8,11 +8,14 @@ See https://github.com/markbates/goth/examples/main.go to see this in action.
 package gothic
 
 import (
+	"encoding/base64"
 	"errors"
 	"fmt"
+	"math/rand"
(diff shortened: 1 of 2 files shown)


markbates:
Thanks Andy!

PR #182, opened 2017-10-08
Moved deferring Logout until it's determined there's actually a sessi…

terev:
Moved deferring of Logout until it's determined there's actually a session to be removed. This fixes an issue encountered when utilizing the redis session store.


terev:
It'd be nice if there were some integration tests utilizing some of the different session stores available.

markbates:
Thanks.

PR #190, opened 2018-01-03
Auth endpoint for auth0 is now /authorize

paganotoni:
I got an error while authenticating with the Auth0 provider and figured is the authorize url.


markbates:
@paganotoni looks like the tests are failing.

PR #194, opened 2018-01-10
Support for AzureAD

michalpristas:
Added support for Azure AD which uses microsoft online and azure ad graph API for fetching user.
Not included in this pack: support for microsoftonline version 2.0

### DIFF
--- a/providers/azuread/azuread.go
+++ b/providers/azuread/azuread.go
@@ -1,0 +1,189 @@
+// Package azuread implements the OAuth2 protocol for authenticating users through AzureAD.
+// This package can be used as a reference implementation of an OAuth2 provider for Goth.
+package azuread
+
+import (
+	"bytes"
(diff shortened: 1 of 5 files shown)


markbates:
I'm having trouble confirming this works, would you mind providing me with some instructions on how to set this up?

Thanks!

michalpristas:
Do you have some azure app set-up? try this tutorial: https://auth0.com/docs/connections/enterprise/azure-active-directory/v2

michalpristas:
i found an issue, please let me get back with an update

markbates:
Thanks!

Any chance of adding support for "normal" logins though MS, not just Azure AD?

michalpristas:
for sure I will look at it over the weekend. Another thing I want to do is in-memory session store so it works with localhost and verification of new providers gets easier

PR #199, opened 2018-01-16
Fix #192 (with the same fix originally in #182)

### LINKED ISSUE #192: Example code not working: gorilla/sessions attempts to delete newly created session.
I'm attempting to get a basic example of google oauth working. As a base case, I tried running the example code provided in goth/examples but upon clicking the "Log in with Google Plus" button it responded with a text based web page saying:
"remove C:\Users\Omar\AppData\Local\Temp\session_: The system cannot find the file specified."

That error comes from a line in gorilla/sessions where the [FilesystemStore.erase](https://github.com/gorilla/sessions/blob/master/store.go#L287) fails to find a file, so my best guess was that there must have been an error saving the session.

I reported this to [gorilla/sessions](https://github.com/gorilla/sessions/issues/137) and they suggested that the package is being misused to get to this state.

I read more about oauth since I'm new to it, and realized I would need to set up a real domain name in order to avoid having to deal with redirect mismatches #94 so I modified the provided demo to:
1. only use gplus as a provider
2. use my own server cert/key for TLS
3. use my own callback url
My code is here: https://gist.github.com/Omustardo/09ae0eb5a3edf3d0d2520d14143d8c46

I've finally had some time to dig into it, and I found out what's happening. 
Looking at the code, the point of failure is on [FilesystemStore.Save](https://github.com/gorilla/sessions/blob/master/store.go#L216) when a session is provided that has Options.MaxAge==-1

The question is, how does it get a session with MaxAge==-1
The session that's causing the error is retrieved by [Store.Get(req, SessionName)](https://github.com/markbates/goth/blob/master/gothic/gothic.go#L284)

Using a debugger, I traced the path of Store.Get and found that the registry it retrieves from contains multiple SessionInfo entries, keyed by session name. In this case, the provided `SessionName` is a constant = "_gothic_session". As far as I can tell, the MaxAge for this session is always -1, so when we then try to save the session, it immediately attempts to erase the local file (at `os.TempDir() + sessions_`) which doesn't yet exist, and so it returns an error.

I suspect I'm missing some key point, since clearly this example has worked for other people (I hope), but it looks to me like the wrong session is being retrieved from `Store.Get`. Looking at the registry, it contains an entry for "_gothic_session" as well as one for "gplus_gothic_session". The gplus one has MaxAge=2592000 which is what we want.

By modifying storeInSession and getFromSession to call `Store.Get(req, key + SessionName)` instead of `Store.Get(req, SessionName)` I was able to get the modified example code to work. I suspect there might be other places in the code where the providerName should be appended to the SessionName but haven't looked deeply yet.

Am I doing something wrong, or should I submit a pull request to fix this?

itay:
We were deleting the session even if one didn't exist. This seems to not work properly with the cookie session store due to multiple calls to Set-Cookie for the same cookie.

### DIFF
--- a/gothic/gothic.go
+++ b/gothic/gothic.go
@@ -147,7 +147,6 @@ as either "provider" or ":provider".
 See https://github.com/markbates/goth/examples/main.go to see this in action.
 */
 var CompleteUserAuth = func(res http.ResponseWriter, req *http.Request) (goth.User, error) {
-	defer Logout(res, req)
 	if !keySet && defaultStore == Store {
 		fmt.Println("goth/gothic: no SESSION_SECRET environment variable is set. The default cookie store is not available and any calls will fail. Ignore this warning if you are using a different store.")
 	}
@@ -167,6 +166,8 @@ var CompleteUserAuth = func(res http.ResponseWriter, req *http.Request) (goth.Us
 		return goth.User{}, err
 	}
 
+	// Only logout to delete the session once we know we have one.
+	defer Logout(res, req)
 	sess, err := provider.UnmarshalSession(value)
 	if err != nil {
 		return goth.User{}, err


markbates:
This was reverted previously because it caused several issues with the cookie store. It also caused a problem with trying to use multiple providers, you would often get set to the wrong provider because the session was still around. 

Another fix has to be found instead of this one.

itay:
Thanks @markbates - appreciate the fast response!

My thinking is that the current state of the library seems mostly non-functional - it doesn't work out of the box with cookie store, and as far as I can tell, doesn't work out of the box with filesystem store either (as you get errors). Given that, it seems better to have it in the previous less broken than the current state, but obviously totally up to you.

markbates:
I use goth in apps all over the place with cookie store and it works perfectly. So if there are issues, please report them. 

As for the file store issue, I’m saying there has to be a fix that fixes it and doesn’t break other stuff. Perhaps checking for the file first? Or checking the error and ignoring if it doesn’t exist. That sort of thing.

itay:
@markbates thanks - I will try and work on a more limited repro than what I have in my app right now.

markbates:
I’m looking for people to take over maintenance of Goth, fyi. In case you’re interested in helping out.

PR #200, opened 2018-01-24
Feature: Slack Provider - Add First and Last Name 

lsiv568:
This PR adds first and last name to the `goth.User` as extracted from slack.  While they are not [guaranteed](https://api.slack.com/methods/users.info) to be present, it's worth trying.


markbates:
Thanks!

PR #205, opened 2018-02-03
Migrated Dropbox endpoints to use v2

### LINKED ISSUE #204: Update providers dropbox to v2
Dropbox Api v1 is no longer supported and simple returns error code 400.
Would you kindly update it to v2?
File using the v1 api: https://github.com/markbates/goth/blob/master/providers/dropbox/dropbox.go
https://www.dropbox.com/developers/reference/migration-guide

adamo57:
* Updates endpoints to point to the v2 of the Dropbox API

Fixes #204

### DIFF
--- a/providers/dropbox/dropbox.go
+++ b/providers/dropbox/dropbox.go
@@ -9,14 +9,15 @@ import (
 	"strings"
 
 	"fmt"
+
 	"github.com/markbates/goth"
 	"golang.org/x/oauth2"
 )
 
 const (
-	authURL    = "https://www.dropbox.com/1/oauth2/authorize"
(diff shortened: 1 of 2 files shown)


bentranter:
Hey @adamo57, thanks for the PR! Looks like Dropbox might've made another change, as when I try to run this, everything works until the point where it tries to get the user in the `FetchUser` function in `dropbox.go`.

It seems like Dropbox now requires a `POST` request for that endpoint, so if you could change https://github.com/adamo57/goth/blob/adamo57-migrate-dropbox-endpoint/providers/dropbox/dropbox.go#L90 to use a `POST` request instead of the current `GET`, then I believe that should fix this.

bentranter:
Awesome, works for me now. Thanks for this!

PR #207, opened 2018-02-20
Export Provider.OpenIDConfig

aileendds:
I'm currently trying to use Goth with an identity provider that uses the OpenID Connect authorization code flow. This flow requires calling the token endpoint. Goth currently extracts the token endpoint URL from auto-discovery, but provides no helpers for calling that endpoint, nor does make the TokenEndpoint field accessible. That means I either need to do my own auto-discovery again, or hardcode the token endpoint URL.

### DIFF
--- a/providers/openidConnect/openidConnect.go
+++ b/providers/openidConnect/openidConnect.go
@@ -54,7 +54,7 @@ type Provider struct {
 	CallbackURL  string
 	HTTPClient   *http.Client
 	config       *oauth2.Config
-	openIDConfig *OpenIDConfig
+	OpenIDConfig *OpenIDConfig
 	providerName string
 
 	UserIdClaims    []string


bentranter:
Hey @aileendds, I'm not sure I fully understand the issue.

It's my understanding that Gothic's [`BeginAuthHandler`](https://github.com/markbates/goth/blob/master/gothic/gothic.go#L61) is the helper for discovering and then calling the token endpoint, but I might be wrong as I'm less familiar with the OpenID Connect spec than I am with the OAuth2 one. There's also the [`GetAuthURL`](https://github.com/markbates/goth/blob/master/gothic/gothic.go#L111) function that contains all the steps for extracting a provider from a request, beginning the auth process, and then creating the auth URL to send users to. Could you have a look through that code to see if there's any functionality there that you could use to accomplish what you're trying to do? If not, then we can look into a different solution.

aileendds:
@bentranter thanks for looking into this! There are different types of OpenID Connect flows. For the "implicit flow", BeginAuthHandler and GetAuthURL would be enough and would return the token. However, my identity provider is using the authorization code flow, which is shown here: https://rograce.github.io/openid-connect-documentation/explore_auth_code_flow.html

In this flow there is a separate request to retrieve the token after the initial one that is handled by BeginAuthHandler / GetAuthURL. I am using BeginAuthHandler to get the authorization code, but unfortunately I need to then hardcode the token URL to make the token request. If you grep through the goth openid connect provider code you can see that TokenEndpoint is set but never used.

PR #210, opened 2018-02-23
Export helpful session access methods

s2t2:
Hi, I'm implementing a custom OIDC provider and found need to use the session access methods `StoreInSession` and `GetFromSession`. I capitalized their function names so they will be exported.

### DIFF
--- a/gothic/gothic.go
+++ b/gothic/gothic.go
@@ -132,7 +132,7 @@ func GetAuthURL(res http.ResponseWriter, req *http.Request) (string, error) {
 		return "", err
 	}
 
-	err = storeInSession(providerName, sess.Marshal(), req, res)
+	err = StoreInSession(providerName, sess.Marshal(), req, res)
 
 	if err != nil {
 		return "", err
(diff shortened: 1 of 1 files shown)


gothic/gothic.go
```diff
@@ -284,7 +284,7 @@ func getProviderName(req *http.Request) (string, error) {
 	return "", errors.New("you must select a provider")
 }
 
-func storeInSession(key string, value string, req *http.Request, res http.ResponseWriter) error {
+func StoreInSession(key string, value string, req *http.Request, res http.ResponseWriter) error {
```
adamo57: please add comment to exported func

gothic/gothic.go
```diff
@@ -294,7 +294,7 @@ func storeInSession(key string, value string, req *http.Request, res http.Respon
 	return session.Save(req, res)
 }
 
-func getFromSession(key string, req *http.Request) (string, error) {
+func GetFromSession(key string, req *http.Request) (string, error) {
```
adamo57: again, add a comment to exported func

s2t2:
OK. Comments added!

Issue #212, opened 2018-02-26
Dropbox provider sets email for goth.User.UserID


kevin-cantwell:
This line looks incorrect: https://github.com/markbates/goth/blob/v1.45.3/providers/dropbox/dropbox.go#L179

```go
user.UserID = u.Email // Dropbox doesn't provide a separate user ID
```

[According to the documentation](https://www.dropbox.com/developers/documentation/http/documentation#users-get_current_account), the correct field to use here from the raw data should be `account_id`.

adamo57:
is this issue fixed @kevin-cantwell

bentranter:
Gonna close this as fixed in #213, we can re-open if it becomes an issue again.

Issue #211, opened 2018-02-26
Dropbox provider does not supply RawData


kevin-cantwell:
Most of the providers in this repo appear to fill in the `goth.User.RawData` field, but not dropbox.

This makes it impossible to parse out the dropbox account id, which is not set as the `goth.User.UserID` either.

bentranter:
I believe this is fixed in the merged PR that is linked. Closing.

PR #213, opened 2018-02-26
fill in all dropbox user details and raw data

kevin-cantwell:
This PR addresses two issues I opened and adds tests:

https://github.com/markbates/goth/issues/211

and

https://github.com/markbates/goth/issues/212

Previously a dropbox user would be missing `RawData` and would also have their `UserID` and `NickName` set to unexpected values (email and first name respectively). This makes sure that the `UserID` comes from the `account_id` field from the raw data response and that `NickName` comes from email (ie: the Dropbox username).

### DIFF
--- a/providers/dropbox/dropbox.go
+++ b/providers/dropbox/dropbox.go
@@ -2,9 +2,11 @@
 package dropbox
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"io"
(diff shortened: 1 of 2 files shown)


bentranter:
Thanks @kevin-cantwell for the contribution! Looks good to me, but I want to give @adamo57 a chance to give this a review before merging.

PR #224, opened 2018-05-24
providers/instagram: use full_name on user.Name

FdeFabricio:
On the Instagram provider, instead of using `username` for both `user.Name` and `user.NickName`, use `full_name` for the `user.Name`.

### DIFF
--- a/providers/instagram/instagram.go
+++ b/providers/instagram/instagram.go
@@ -130,7 +130,7 @@ func userFromReader(reader io.Reader, user *goth.User) error {
 		return err
 	}
 	user.UserID = u.Data.ID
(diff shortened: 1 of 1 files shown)


bentranter:
Looks good to me, using the full name for an identity check (which is what we do) makes more sense than using their username.

Relevant docs for anyone else who sees this: https://www.instagram.com/developer/endpoints/users/#get_users_self (you have to click the "Response" button to actually see the response in the docs -- took me a minute to figure out how to see that 😛).

bentranter:
Thanks for the contribution @FdeFabricio!! 🎉

PR #226, opened 2018-07-03
[cf] Each oauth2 access token has a token type, typically 'bearer'; need to return this for CF UAA

drnic:


drnic:
closing. AFAIK its always bearer.

PR #227, opened 2018-07-03
[cf] store CF UAA authorized scopes in RawData['scopes']

drnic:


drnic:
I'll close this aged PR now. If it solves someone's problem in future we can reopen it.

PR #240, opened 2018-09-06
Follow the OpenID Connect Audiences spec

OvermindDL1:
This follows the OpenID Connect spec to support audiences with either a string or an array of strings instead of just a string.

This fixes the bug at https://github.com/go-gitea/gitea/issues/4877 and the PR at https://github.com/go-gitea/gitea/pull/4878 is dependent on this getting fixed and released.

I have tested this and am running it on my site now with the OpenID Connect with multiple audiences returned with no issue, the new code is not even run at all unless the old pure-string check fails.

Disclaimer: I've never used and barely looked at Go thus far, and I was unable to find common functions like filter or so forth (or any higher order functions at all...) and based on some googling it appears that these large and unwieldy loops are the standard way to do this kind of thing in Go, but if any changes are needed please request as such with information on what I should change it to (great learning opportunity!).

And of course any rights or whatever there-of related to this code I give up for whatever use this project currently wants or will want in the future. Etc... etc... Commit is signed, etc...

### DIFF
--- a/providers/openidConnect/openidConnect.go
+++ b/providers/openidConnect/openidConnect.go
@@ -200,7 +200,17 @@ func (p *Provider) RefreshToken(refreshToken string) (*oauth2.Token, error) {
 func (p *Provider) validateClaims(claims map[string]interface{}) (time.Time, error) {
 	audience := getClaimValue(claims, []string{audienceClaim})
 	if audience != p.ClientKey {
-		return time.Time{}, errors.New("audience in token does not match client key")
+		found := false
+		audiences := getClaimValues(claims, []string{audienceClaim})
+		for _, aud := range audiences {
+			if aud == p.ClientKey {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return time.Time{}, errors.New("audience in token does not match client key")
+		}
 	}
 
 	issuer := getClaimValue(claims, []string{issuerClaim})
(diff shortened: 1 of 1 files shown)


OvermindDL1:
For note, the OpenID Connect Spec URL that is actually referenced in this file (4 lines above the first diff line actually) confirms that the audience can be a string or an array and both must be handled, hence why this is not working on the OpenID Connect server that I have to use since it sends multiple audiences as an array of strings.

A quoted summary from https://github.com/go-gitea/gitea/issues/4877

> At this line of code:
> https://github.com/markbates/goth/blob/ed4f7378f4655f3c2200f93627e7edcf223b1445/providers/openidConnect/openidConnect.go#L202
> ```go
> 	if audience != p.ClientKey {
> ```
> It is performing a string comparison between `audience` and `p.ClientKey`, however, according to the OpenID Connect spec that that very source file links to just 3 lines higher states (emphasis mine):
> 
> > The Client MUST validate that the `aud` (audience) Claim contains its `client_id` value registered at the Issuer identified by the `iss` (issuer) Claim as an audience. ***The `aud` (audience) Claim MAY contain an array with more than one element.*** The ID Token MUST be rejected if the ID Token does not list the Client as a valid audience, or if it contains additional audiences not trusted by the Client. 
> 
> And in fact an array is being passed in by some OpenID Connect servers, which is causing this to fail and be unusable.

OvermindDL1:
Is there anything else that I need to do to get this merged in and released?

bentranter:
Sorry for the delay @OvermindDL1, I was on vacation when this initially came in and missed it in a sea of other notifications 😛

I'd say this looks good to me, and while I'd like to add a test for this so it doesn't get broken in the future, this is great for now, and I really appreciate the amount of detail put into both the comments -- made it very easy to review. Thanks for the contribution!

OvermindDL1:
> Sorry for the delay @OvermindDL1, I was on vacation when this initially came in and missed it in a sea of other notifications stuck_out_tongue

All good, just making sure I didn't miss anything.  :-)

>  I'd say this looks good to me, and while I'd like to add a test for this so it doesn't get broken in the future, this is great for now, and I really appreciate the amount of detail put into both the comments -- made it very easy to review. Thanks for the contribution!

Thanks!  I do apologize for the lack of tests, traditionally I would in C++ or so but I'm entirely new to Go so this PR took me an excess of time as it was as I kept trying to look up patterns that Go wasn't capable of...  ^.^;

PR #242, opened 2018-10-01
Added support for Tumblr

BakeRolls:
Added support for Tumblr. This is basically a working copy of another provider with modified URLs. Please let me know if I have to change something else.


adamo57:
Looks good to me

PR #244, opened 2018-10-08
Remove maintainer request from README

cippaciong:
Since Goth has a new maintainer, that part of the README is misleading and can lead new users to think that is unmaintained.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,10 +8,6 @@ protocol providers, as long as they implement the `Provider` and `Session` inter
 
 This package was inspired by [https://github.com/intridea/omniauth](https://github.com/intridea/omniauth).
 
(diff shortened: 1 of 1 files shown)


bentranter:
This is probably a good idea, curious as to what @adamo57 thinks too though.

cippaciong:
Sure. Just for clarity, this was briefly discussed on #buffalo Slack channel before I made the PR. But of course it's up to the current maintainer to decide what to do :+1:

bentranter:
Oh cool, I hadn't noticed that, although with how active the Buffalo channel is it's easy to miss things 😛 I'll merge this based on that discussion then, thanks for pointing that out.

PR #245, opened 2018-10-12
Add simple Google provider

### LINKED ISSUE #214: Simple `auth/google` provider needed in addition to gplus
Currently `gplus` is the only Google provider available for goth.   A simple google api is also needed for applications where Google+ features or Google+ accounts may not exist.

From the Google+ SignIn docs...   (https://developers.google.com/+/web/signin/)

*"Note: If you only want to sign in your users with their Google Accounts, and do not need to use any Google+ features, integrate Google Sign-In into your app without requesting the Google+ scopes as described below."*

* https://developers.google.com/identity/sign-in/web/

bentranter:
### Overview

Adds a simple Google provider. Closes #214.

### Summary

Adds a new "Google" provider. This essentially clones the existing `gplus` provider, with a few differences:

* It will, by default, only send the `email` scope, since `profile` is a Google+ specific scope, and `openid` is a bit excessive for a default scope.
* It uses the `oauth2.Endpoint` from `golang.org/x/oauth2/google` so we always keep up to date with Go's OAuth2 package.
* It simplifies the logic that populates a `goth.User` from the JSON returned from Google's server.

It also includes a few housekeeping items:

1. Marks the Google+ provider as deprecated in the README.
2. Updates the `go.mod` file.
3. Checks in the `go.sum` file. I know not everyone does this, but for now it seems to be community consensus that it's a good idea to check it in, so I figure we'd do that.


bentranter:
I'll let this sit over the weekend and Monday, and if there's no feedback by Tuesday, I'll merge it as-is to resolve #214.

Issue #236, opened 2018-08-20
Option for changing facebook image size


kobayashi:
Retrieved facebook's image is small size (height: 50px, width: 50px).
Is there an option to change the retrieved image size?

[Facebook Graph API ](https://developers.facebook.com/docs/graph-api/reference/user/picture/) provides multiple image urls by specifying picture type.

kobayashi:
Added this feature by #263

PR #246, opened 2018-10-20
Add facebook field expansion support

filinvadim:
#236


bentranter:
Looks good, we've needed this for a while. Thanks a ton @filinvadim!

bubenkoff:
this introduced an issue with FB giving 500 error
screen shot 2018-11-19 at 17 41 23
because we now pass invalid values in the scope like first_name, last_name etc -> those are not the valid scope values, those are fields, this is not the same thing

PR #247, opened 2018-10-24
Migrate x/net/context to standard lib context

sapk:
Since Go 1.7 this package is available in the standard library under the name context.  https://golang.org/pkg/context.

Looking at the travis CI, I found that you support only Go1.7+ so this will permit to use the standard definition of context.

### DIFF
--- a/provider.go
+++ b/provider.go
@@ -1,10 +1,10 @@
 package goth
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 
-	"golang.org/x/net/context"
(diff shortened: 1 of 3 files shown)


bentranter:
Thanks for this @sapk, looks good to me.

I considered asking if you could run `go mod tidy` to remove the dependency on `x/net/context` in the `go.mod` and `go.sum` files, but it looks like our `go.mod` and `go.sum` files are a bit out of date anyway, so I'll update them myself in a separate PR.

PR #250, opened 2018-11-17
Rearrange handlers blocks

emad-elsaid:
This will rearrange the handlers blocks to have a more natural order, reflecting the user journey through the pages

1. root 
2. auth
3. callback
4. logout


emad-elsaid:
I see now why this won't work with `pat` as it depends on that order

PR #254, opened 2018-12-06
Reject Twitch users with unverified emails

Prendo93:
Twitch will return oauth info including emails for users that have just signed up and have not yet verified their emails.

Trusting this information caused a vulnerability in our app, and could likely do the same for a naiive user of the library. 

Unsure if this meets the standards for a patch, but thought it worth raising as a discussion point.

### DIFF
--- a/providers/spalktestprovider/spalktestprovider.go
+++ b/providers/spalktestprovider/spalktestprovider.go
@@ -1,0 +1,159 @@
+package spalktestprovider
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	"github.com/markbates/goth"
+	"golang.org/x/oauth2"
+)
+
+// Provider is used only for testing.
+type Provider struct {
+	HTTPClient   *http.Client
+	providerName string
+}
+
+// Session is used only for testing.
+type Session struct {
+	ID          string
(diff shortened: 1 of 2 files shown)


bentranter:
@Prendo93 Hmm, I'm not sure about this. It looks like the behaviour in the old version of the API was not to return the user's email (see [the discussion here](https://discuss.dev.twitch.tv/t/oauth-knowing-whether-a-user-has-verified-their-email/16505)), and with the new Helix API, they return the `email_verified` bool. However, we still use the old API endpoints, so I'm not sure if the `email_verified` value is even set on the response.

While I can't merge your PR (you imported a dependency from a private repo), I'd consider creating a new provider for the Twitch Helix API to support this behaviour.

I'm gonna do a bit more research and look at the new Twitch Helix API, and I'll come back to this sometime this weekend. Until then, I'd love to hear if anyone else has an opinion on how to handle this.

Prendo93:
@bentranter thanks for the quick response. I figured our hasty patch wouldn't be accepted but wanted to make the maintainers aware. If this is something the project is interested in, I'm happy to fix up deps and add tests for the patch.

We are also looking at making the 

FWIW, our experience of the API is that even using the old (`/kraken`) endpoints set at the top of `twitch.go`, the `email` and `email_verified: false` fields are both set on the struct when the account is not verified, and the `email_verified` is absent when the account is verified.

PR #256, opened 2018-12-10
updated battle.net provider to use the new oauth serivces

Toyz:
I fixed the battle.net provider to use the new system from https://develop.battle.net/ as us.api.battle.net is no longer used for userinfo in oauth

### DIFF
--- a/providers/battlenet/battlenet.go
+++ b/providers/battlenet/battlenet.go
@@ -85,7 +85,11 @@ func (p *Provider) FetchUser(session goth.Session) (goth.User, error) {
 
 	// Get the userID, battlenet needs userID in order to get user profile info
 	c := p.Client()
-	req, _ := http.NewRequest("GET", endpointUser, nil)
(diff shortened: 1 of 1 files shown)


providers/battlenet/battlenet.go
```diff
@@ -84,7 +84,11 @@ func (p *Provider) FetchUser(session goth.Session) (goth.User, error) {
 	}
 
 	// Get the userID, battlenet needs userID in order to get user profile info
-	response, err := p.Client().Get(endpointUser + "?access_token=" + url.QueryEscape(sess.AccessToken))
+	c := p.Client()
+	req, _ := http.NewRequest("GET", endpointUser, nil)
```
bentranter: The error here should be handled, not ignored.

Toyz:
I just updated this to add error handling, sorry about that was a rushed fix for my site i forgot to add that for the PR

PR #262, opened 2019-01-03
google provider should populate user.RawData (as gplus did)

danielbprice:
Changes the `google` provider to populate `user.RawData`, which the gplus provider did.  There are some important fields to be had, such as `hd`.

Also moves the call to defer response.Body.Close() earlier.

I don't do a lot of pull requests, so apologies if I've screwed some aspect up.


bentranter:
Looks good to me, thanks @danielbprice!

Issue #252, opened 2018-11-23
Facebook oauth unable to remove scope


andrearizzello:
We are doing a project where a user can authenticate through facebook, the problem is that we don't know how to remove the scope, since we don't need it.
We need to remove this `&scope=email+first_name+last_name+link+about+id+name+picture+location` from the url

bentranter:
This sounds related to the comment in https://github.com/markbates/goth/pull/246.

If you could open a pull request for this @andrearizzello, that would be much appreciated, as I have no Facebook account and am not able to manually test them as an authentication provider.

PR #263, opened 2019-01-03
add fields options

kobayashi:
This PR is for issue #236, #252. 

PR #246 does not work correctly because of inappropriate scopes.

The following is a sample code to add custom fields.

```
func init() {
	gothic.Store = App().SessionStore
	query := []string{"email", "first_name", "last_name", "id", "name", "picture.type(large)", "location"}
	goth.UseProviders(
		facebook.New(os.Getenv("FACEBOOK_KEY"), os.Getenv("FACEBOOK_SECRET"), fmt.Sprintf("%s%s", "http://localhost:3000", "/auth/facebook/callback")).SetCustomFields(query),

	)
}
```

### DIFF
--- a/providers/facebook/facebook.go
+++ b/providers/facebook/facebook.go
@@ -62,8 +62,11 @@ func (p *Provider) SetName(name string) {
 	p.providerName = name
 }
 
-// SetCustomFields is to update the fields for endpointProfile
-// Fields values are referd with https://developers.facebook.com/docs/graph-api/reference/user
+// SetCustomFields sets the fields used to return information
+// for a user.
+//
+// A list of available field values can be found at
+// https://developers.facebook.com/docs/graph-api/reference/user
 func (p *Provider) SetCustomFields(fields []string) *Provider {
 	p.Fields = strings.Join(fields, ",")
 	return p
--- a/providers/facebook/facebook.go
+++ b/providers/facebook/facebook.go
@@ -37,6 +37,7 @@ func New(clientKey, secret, callbackURL string, scopes ...string) *Provider {
 		providerName: "facebook",
 	}
 	p.config = newConfig(p, scopes)
+	p.Fields = "email,first_name,last_name,link,about,id,name,picture,location"
 	return p
 }
 
@@ -46,6 +47,7 @@ type Provider struct {
 	Secret       string
 	CallbackURL  string
 	HTTPClient   *http.Client
+	Fields       string
 	config       *oauth2.Config
 	providerName string
 }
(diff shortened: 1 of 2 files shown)


providers/facebook/facebook.go
```diff
@@ -60,6 +62,13 @@ func (p *Provider) SetName(name string) {
 	p.providerName = name
 }
 
+// SetCustomFields is to update the fields for endpointProfile
+// Fields values are referd with https://developers.facebook.com/docs/graph-api/reference/user
```
bentranter: This comment is a bit hard to understand. It should be updated to say something like,

```go
// SetCustomFields sets the fields used to return information
// for a user.
//
// A list of available field values can be found at
// https://developers.facebook.com/docs/graph-api/reference/user
```

providers/facebook/facebook.go
```diff
@@ -37,6 +37,7 @@ func New(clientKey, secret, callbackURL string, scopes ...string) *Provider {
 		providerName: "facebook",
 	}
 	p.config = newConfig(p, scopes)
+	p.Fields = "email,first_name,last_name,link,about,id,name,picture,location"
```
bentranter: In the interest of avoiding any more issues with the Facebook provider in the future, I'm wondering, what is the relationship between a "scope" and a "field" when it comes to permissions? Do the scopes match one-to-one with the fields? For example, if I authenticate with the scope "email", am I allowed to see the user's name and location, or would I need additional scopes for that?

providers/facebook/facebook.go
kobayashi: OK. I will change the comment.

providers/facebook/facebook.go
kobayashi: Yeah, that makes us confused. Parameters of scopes are not same with those of fields. As you know, "scope" is used for user permissions, what type of information is permitted to access by the token. Easiest way to try to find which permissions are available is at [Graph API Explorer](https://developers.facebook.com/tools/explorer/).
On the other hand, "fields" is for which actual field you want to get, post, delete.

Here are some examples.
- When I get a token with scope "email", it can get "id", "name", "email". "id" and "name" are default parameters in Graph api v3.2.
- If I want to get users' profile links, I have to set "user_link" in scopes. Then I can retrieve the link by setting fields="link".

bentranter: Ok, thanks, I think I get it. Scopes manage permissions, and the fields work in a sort of GraphQL way -- you can request specific data via the scopes.

I'm happy to merge this, hopefully this resolves the issues. Thanks for the contribution @kobayashi!

PR #265, opened 2019-01-08
Add nextcloud

lnxbil:
I adopted the Gitlab provider to work with a self-hosted Nextcloud instance and also tried to describe  the setup of a test environment for performing the tests with a self-hosted Docker Image. Program and tests are normally more complex, because we have a dynamic endpoint instead of static compiled in one.

I also tried to simplify the setup and introduced a new `New` method named `NewCustomisedDNS` that only needs one customised variable in order to work.


providers/nextcloud/nextcloud.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	if user.AccessToken == "" {
+		// data is not yet retrieved since accessToken is still empty
+		return user, fmt.Errorf("%s cannot get user information without accessToken", p.providerName)
+	}
+
+	req, err := http.NewRequest("GET", p.profileURL, nil)
+	if err != nil {
+		fmt.Println(err)
```
bentranter: This `fmt.Println` is unnecessary because the error is returned, so it should be removed.

providers/nextcloud/nextcloud.go
```diff
@@ -0,0 +1,211 @@
...
+		return user, err
+	}
+	req.Header.Set("Authorization", "Bearer "+sess.AccessToken)
+	response, err := p.Client().Do(req)
+
+	if err != nil {
+		if response != nil {
+			response.Body.Close()
+		}
+		fmt.Println(err)
```
bentranter: This `fmt.Println` is unnecessary because the error is returned, so it should be removed.

providers/nextcloud/nextcloud.go
```diff
@@ -0,0 +1,211 @@
...
+
+	defer response.Body.Close()
+
+	if response.StatusCode != http.StatusOK {
+		return user, fmt.Errorf("%s responded with a %d trying to fetch user information", p.providerName, response.StatusCode)
+	}
+
+	bits, err := ioutil.ReadAll(response.Body)
+	if err != nil {
+		fmt.Println(err)
```
bentranter: This `fmt.Println` is unnecessary because the error is returned, so it should be removed.

providers/nextcloud/nextcloud.go
```diff
@@ -0,0 +1,211 @@
...
+		return user, err
+	}
+
+	err = json.NewDecoder(bytes.NewReader(bits)).Decode(&user.RawData)
+	if err != nil {
+		fmt.Println(err)
```
bentranter: This `fmt.Println` is unnecessary because the error is returned, so it should be removed.

providers/nextcloud/nextcloud.go
```diff
@@ -0,0 +1,211 @@
...
+		return user, err
+	}
+
+	err = userFromReader(bytes.NewReader(bits), &user)
+
+	fmt.Println(err)
```
bentranter: This `fmt.Println` is unnecessary because the error is returned, so it should be removed.

providers/nextcloud/nextcloud.go
```diff
@@ -0,0 +1,211 @@
...
+				EMail       string `json:"email"`
+				DisplayName string `json:"display-name"`
+				ID          string `json:"id"`
+				Address     string `json:"address"`
+			}
+		} `json:"ocs"`
+	}{}
+	err := json.NewDecoder(r).Decode(&u)
+	if err != nil {
+		fmt.Println(err)
```
bentranter: This `fmt.Println` is unnecessary because the error is returned, so it should be removed.

providers/nextcloud/nextcloud.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+	return c
+}
+
+func userFromReader(r io.Reader, user *goth.User) error {
+	u := struct {
+		Ocs struct {
+			Data struct {
+				EMail       string `json:"email"`
+				DisplayName string `json:"display-name"`
```
bentranter: Does NextCloud really return the the JSON payload with the key `"display-name"` (not `"display_name"`)? That's a bit odd to me, so I'm just wondering if this is a typo or if they really do use that key.

providers/nextcloud/nextcloud.go
```diff
@@ -0,0 +1,211 @@
...
+func (p *Provider) RefreshTokenAvailable() bool {
+	return true
+}
+
+//RefreshToken get new access token based on the refresh token
+func (p *Provider) RefreshToken(refreshToken string) (*oauth2.Token, error) {
+	token := &oauth2.Token{RefreshToken: refreshToken}
+	ts := p.config.TokenSource(goth.ContextForClient(p.Client()), token)
+	newToken, err := ts.Token()
+	if err != nil {
```
bentranter: This can be simplified to just be,

```go
return ts.Token()
```

providers/nextcloud/nextcloud.go
```diff
@@ -0,0 +1,211 @@
...
+		return err
+	}
+	user.Email = u.Ocs.Data.EMail
+	user.Name = u.Ocs.Data.DisplayName
+	user.UserID = u.Ocs.Data.ID
+	user.Location = u.Ocs.Data.Address
+	return nil
+}
+
+//RefreshTokenAvailable refresh token is provided by auth provider or not
```
bentranter: There should be a space between the comment slashes `//` and the start of the comment. Also, this sentence is a bit strange to read, it should be something more like,

```
// RefreshTokenAvailable returns true if a refresh token is provided by the auth
// provider, and false if not.
```

providers/nextcloud/nextcloud.go
```diff
@@ -0,0 +1,211 @@
...
+func (p *Provider) RefreshTokenAvailable() bool {
+	return true
+}
+
+//RefreshToken get new access token based on the refresh token
```
bentranter: Same issue here as above, and one small grammar nit: it should be,

```go
// RefreshToken gets ...
```

instead of "get".

providers/nextcloud/session_test.go
```diff
@@ -0,0 +1,48 @@
+package nextcloud_test
+
+import (
+	"testing"
+
+	"github.com/markbates/goth"
+	"github.com/markbates/goth/providers/gitlab"
```
bentranter: This should import the `nextcloud` package, not `gitlab`.

providers/nextcloud/session_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_Implements_Session(t *testing.T) {
+	t.Parallel()
+	a := assert.New(t)
+	s := &gitlab.Session{}
```
bentranter: Should be `nextcloud`, not `gitlab`.

providers/nextcloud/session_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	a.Implements((*goth.Session)(nil), s)
+}
+
+func Test_GetAuthURL(t *testing.T) {
+	t.Parallel()
+	a := assert.New(t)
+	s := &gitlab.Session{}
```
bentranter: Should be `nextcloud`, not `gitlab`.

providers/nextcloud/session_test.go
```diff
@@ -0,0 +1,48 @@
...
+	s.AuthURL = "/foo"
+
+	url, _ := s.GetAuthURL()
+	a.Equal(url, "/foo")
+}
+
+func Test_ToJSON(t *testing.T) {
+	t.Parallel()
+	a := assert.New(t)
+	s := &gitlab.Session{}
```
bentranter: Should be `nextcloud`, not `gitlab`.

providers/nextcloud/session_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+	data := s.Marshal()
+	a.Equal(data, `{"AuthURL":"","AccessToken":"","RefreshToken":"","ExpiresAt":"0001-01-01T00:00:00Z"}`)
+}
+
+func Test_String(t *testing.T) {
+	t.Parallel()
+	a := assert.New(t)
+	s := &gitlab.Session{}
```
bentranter: Should be `nextcloud`, not `gitlab`.

bentranter:
@lnxbil There are still a few outstanding comments to address, mostly the ones around removing unnecessary `fmt.Println` statements and changing the function names as I suggested. After those are addressed, I'll merge.

providers/nextcloud/nextcloud.go
lnxbil: Yes, they really do. The API is a bit tricky to find in their code, but [this test does expect it](https://github.com/nextcloud/server/blob/42121fe4d71237f769831e2b33def779e0566e7c/apps/provisioning_api/tests/Controller/UsersControllerTest.php#L3056).

lnxbil:
Oh darn ... the other comments were folded away. Sorry.

Hmm ... this folding seems to be a GitHub feature (or just pure luck) that relates the code to already present code, because your comments are with respect to lines that are identical to the GitLab implementation I copied from, including the spelling of `customise`, which is the only spelling form of that word in the repository, I checked, because I normally also write it the "north-american way".

```
$ grep -ir customize . | wc -l
       0
$ grep -ir customise . | wc -l
      50
```

and is also present in the `github`, `paypal` and `influxcloud` provider.

So, what should we do here? Stick to the spelling currently used? I don't think that the packages already using the other providers will be happy if we break the API or introduce a different spelling.

What about the other issues like code simplification, comment improvement (get vs. gets) etc., that also applies to the source, the `gitlab` provider?

In the mean time, I'll work on the other comments.

BTW: Thank you for giving me so much feedback, it is great to work on this project.

bentranter:
@lnxbil Thanks for making the changes 🙂

I wasn't aware that the other provider packages used that spelling as well -- thanks for looking through them. We might as well keep it consistent with those packages, as consumers of this library are more likely to be expecting the spelling you added rather than the spelling I suggested.

I'd say this is good to merge now. Thanks for the contribution, and and extra thank you for such thorough documentation on Nextcloud!

lnxbil:
I'm going to try to simplify the code, I just reviewed the provider interface and realized, that a lot of the copy&pasted stuff is unnecessary. I'll report back.

lnxbil:
a few seconds too late. I'll open another PR

bentranter:
@lnxbil Sounds great, I'll hold off on creating a new release until you open that next PR.

PR #266, opened 2019-01-28
add log line about which port is being listened on to example

apuchitnis:
this makes using the example easier


apuchitnis:
ptal @bentranter :)

PR #272, opened 2019-03-26
Do not use user_ids field in VK provider

lukashes:
Because user_ids is an optional field and it does not allow
to use VK provider with already known access_token


bentranter:
Looks good to me, thanks @lukashes for the contribution!

PR #273, opened 2019-04-02
Added mailru provider

snovichkov:
Please review my PR. Thx.

### DIFF
--- a/providers/mailru/mailru.go
+++ b/providers/mailru/mailru.go
@@ -1,0 +1,165 @@
+// Package mailru implements the OAuth2 protocol for authenticating users through mailru.com.
+// This package can be used as a reference implementation of an OAuth2 provider for Goth.
+package mailru
+
+import (
+	"crypto/md5"
+	"encoding/json"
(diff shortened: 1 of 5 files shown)


providers/mailru/mailru.go
```diff
@@ -0,0 +1,166 @@
...
+			Provider:     p.Name(),
+			ExpiresAt:    sess.ExpiresAt,
+		}
+	)
+
+	if user.AccessToken == "" {
+		return user, fmt.Errorf("%s cannot get user information without access token", p.name)
+	}
+
+	var builder strings.Builder
```
bentranter: `strings.Builder` is only available for Go 1.10 and later, but Goth supports back to Go 1.7. You'll need to either guard this with a build tag, or to make things easier, use some other means of creating this string. String concatenation is probably the way to go.

snovichkov: Oops, fixed.

bentranter:
Thanks for the contribution @snovichkov!

PR #275, opened 2019-05-06
Added LINE Provider

rbo13:
- This PR adds the LINE as a new provider for this package.

### DIFF
--- a/providers/line/line.go
+++ b/providers/line/line.go
@@ -1,3 +1,157 @@
 // Package line implements the OAuth2 protocol for authenticating users through line.
 // This package can be used as a reference implementation of an OAuth2 provider for Goth.
 package line
+
+import (
+	"bytes"
+	"encoding/json"
+	"io/ioutil"
+	"net/http"
+
+	"fmt"
+
+	"github.com/markbates/goth"
+	"golang.org/x/oauth2"
+)
+
+const (
+	authURL      string = "https://access.line.me/oauth2/v2.1/authorize"
+	tokenURL     string = "https://api.line.me/oauth2/v2.1/token"
+	endpointUser string = "https://api.line.me/v2/profile"
+)
+
+// Provider is the implementation of `goth.Provider` for accessing Battle.net.
+type Provider struct {
+	ClientKey    string
+	Secret       string
+	CallbackURL  string
+	HTTPClient   *http.Client
+	config       *oauth2.Config
+	providerName string
+}
+
+// New creates a new Line provider and sets up important connection details.
+// You should always call `line.New` to get a new provider.  Never try to
+// create one manually.
+func New(clientKey, secret, callbackURL string, scopes ...string) *Provider {
+	p := &Provider{
+		ClientKey:    clientKey,
+		Secret:       secret,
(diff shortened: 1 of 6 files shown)


rbo13:
Please see my work @bentranter CC @markbates. Thank you so much, would love to hear your feedback since its my first time contributing to open-source project 🤓🥳. Awesome project, btw 🎉

bentranter:
Looks good @rbo13, thanks for the contribution! Could you also add your new provider to the example at https://github.com/markbates/goth/blob/master/examples/main.go? After, that, I will happily merge this 🙂

rbo13:
Sure thing @bentranter 👍 will update this PR :)

rbo13:
Please don't merge this yet @bentranter, seems I have trouble getting user information(e.g User Name) in LINE.


Screen Shot 2019-05-07 at 09 53 38

rbo13:
This PR: https://github.com/markbates/goth/pull/275/commits/b458a55d00668853bbf19cbf145b6a3412fd59eb should fix the issue @bentranter 👍

rbo13:
Oops, accidentally closed my PR :( will re-open.

providers/line/line.go
```diff
@@ -0,0 +1,162 @@
...
+		DisplayName   string `json:"displayName"`
+		PictureURL    string `json:"pictureUrl"`
+		StatusMessage string `json:"statusMessage"`
+	}{}
+
+	if err = json.NewDecoder(bytes.NewReader(bits)).Decode(&u); err != nil {
+		return user, err
+	}
+
+	name := strings.Split(u.DisplayName, " ")
```
rbo13: This line needs an improvement, what do you think @bentranter? Don't have any idea how to properly handle splitting strings though 😅. From the docs of [LINE](https://developers.line.biz/en/docs/social-api/getting-user-profiles/) the display name is a one line string, so I need to split it to get the First and Last Name.

bentranter: This seems ok to me, although I'm worried about the possible case where `u.DisplayName` is only one word long, so that `strings.Split(u.DisplayName, " ")` returns a slice containing a single string. In that case, trying to access `name[1]` would panic.

It might be simpler not to bother setting the first and last name (since LINE doesn't seem to support those fields explicitly), and just setting the `u.DisplayName` to the nickname field. I also wouldn't bother changing the display name to lowercase before setting it to nickname, in case a user of this provider wants to compare display names as-is.

rbo13: > returns a slice containing a single string. In that case, trying to access name[1] would panic.

- Yeah, it will really panic. But now, I won't bother setting it :) thanks for the feedback 👍 

> I also wouldn't bother changing the display name to lowercase before setting it to nickname, in case a user of this provider wants to compare display names as-is.

- On it. Will update my PR to apply all the reviewed changes :)

Thanks @bentranter 🍻

rbo13:
Fixes has been applied @bentranter 👍 please check it if you have time 😅. Thank you so much! Cheers to this amazing library 🍻

rbo13:
Thanks @bentranter 👍 I'm so happy that my first contribution to open-source was accepted and merged 😅.

PR #276, opened 2019-05-16
Add Gitea as a Provider

techknowlogick:


providers/gitea/gitea.go
```diff
@@ -0,0 +1,186 @@
...
+func (p *Provider) Debug(debug bool) {}
+
+// BeginAuth asks Gitlab for an authentication end-point.
+func (p *Provider) BeginAuth(state string) (goth.Session, error) {
+	return &Session{
+		AuthURL: p.config.AuthCodeURL(state),
+	}, nil
+}
+
+// FetchUser will go to Gitlab and access basic information about the user.
```
bentranter: Looks like there's a small copy-paste error here in the doc string – it should say `gitea` instead of `Gitlab`.

providers/gitea/gitea.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+func (p *Provider) Client() *http.Client {
+	return goth.HTTPClientWithFallBack(p.HTTPClient)
+}
+
+// Debug is a no-op for the gitea package.
+func (p *Provider) Debug(debug bool) {}
+
+// BeginAuth asks Gitlab for an authentication end-point.
```
bentranter: Also here, doc string should say `gitea` instead of `Gitlab`

providers/gitea/session.go
```diff
@@ -0,0 +1,63 @@
...
+import (
+	"encoding/json"
+	"errors"
+	"strings"
+	"time"
+
+	"github.com/markbates/goth"
+)
+
+// Session stores data during the auth process with Gitlab.
```
bentranter: Also here, doc string should say `gitea` instead of `Gitlab`

providers/gitea/session.go
```diff
@@ -0,0 +1,63 @@
...
+type Session struct {
+	AuthURL      string
+	AccessToken  string
+	RefreshToken string
+	ExpiresAt    time.Time
+}
+
+var _ goth.Session = &Session{}
+
+// GetAuthURL will return the URL set by calling the `BeginAuth` function on the Gitlab provider.
```
bentranter: Also here, doc string should say `gitea` instead of `Gitlab`

providers/gitea/session.go
```diff
@@ -0,0 +1,63 @@
...
+func (s Session) GetAuthURL() (string, error) {
+	if s.AuthURL == "" {
+		return "", errors.New(goth.NoAuthUrlErrorMessage)
+	}
+	return s.AuthURL, nil
+}
+
+// Authorize the session with Gitlab and return the access token to be stored for future use.
```
bentranter: Also here, doc string should say `gitea` instead of `Gitlab`

techknowlogick:
Thanks for catching those items @bentranter. I've resolved the copy/paste items and also added an entry to the README.

PR #277, opened 2019-05-16
Update travis to use go 1.12

techknowlogick:


bentranter:
Whoops, we really should've added 1.12 to the matrix back when it was released – thanks a ton @techknowlogick for adding this!

PR #282, opened 2019-06-17
Make Salesforce Provider URLs user updatable

### LINKED ISSUE #280: salesforce provider doesn't support salesforce community
I was trying to set up SSO for salesforce community using salesforce provider. Goth has hardcoded links for authorization and token endpoint. There should be a way to specify community links.

As salesforce now does support openid-connect flow, a possible fix could be we follow the same workflow as openid-connect provider. Instead of using value of `subject` field as userId, fetch value of `user_id` field of UserInfo endpoint.

kkirsche:
Fix #280 

@komalsukhani if you can provide an example URL structure, I can update the comment to be more accurate

### DIFF
--- a/providers/salesforce/salesforce.go
+++ b/providers/salesforce/salesforce.go
@@ -13,9 +13,15 @@ import (
 	"golang.org/x/oauth2"
 )
 
-const (
-	authURL  string = "https://login.salesforce.com/services/oauth2/authorize"
-	tokenURL string = "https://login.salesforce.com/services/oauth2/token"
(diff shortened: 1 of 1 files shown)


kkirsche:
Amended previous commit with fixes for failed build in CI

bentranter:
Looks good, thanks for the contribution @kkirsche!

komalsukhani:
@kkirsche For community, we just need to replace host name with community url.
 So basic structure for URLs will be 
**Auth URL** : https://<community_url>/services/oauth2/authorize
**Token URL** : https://<community_url>/services/oauth2/token

PR #286, opened 2019-08-11
Fix provider deduction from existing sessions so that an argument is not needed on the authentication landing page.

akramer:
This has been broken since https://github.com/markbates/goth/commit/78433fec1a8fed110af91e79c94be2d835afe637.

Also moving the deduction code to be a fallback: if a provider is explicitly
specified it should not be overridden by the existing session. I believe
this is why the Logout() problems with multiple providers existed in the
first place. If deduction wins over explicit parameters, then once you
wrongly click one provider, the cookie will force you to use that
authentication mechanism even if you've navigated to a URL that
explicitly states another provider's name.

Fixed the test harness to check session key names (this was a bug in the
test harness) and added a test to verify that this change works.

### DIFF
--- a/gothic/gothic.go
+++ b/gothic/gothic.go
@@ -245,19 +245,6 @@ var GetProviderName = getProviderName
 
 func getProviderName(req *http.Request) (string, error) {
 
-	// get all the used providers
-	providers := goth.GetProviders()
-
-	// loop over the used providers, if we already have a valid session for any provider (ie. user is already logged-in with a provider), then return that provider name
-	for _, provider := range providers {
-		p := provider.Name()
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/markbates/goth
+module github.com/akramer/goth
```
bentranter: Looks like this change was probably added by accident when you worked on your fork – could you remove this?

akramer:
Ugh - I didn't realize that pushing to master would automatically update the pull request. Live and learn. :)

I've reverted the go.mod change.

bentranter:
Thanks @akramer. Going to merge this, hopefully there're no further issues.

Issue #290, opened 2019-09-04
Add support for Apple Oauth


tylerstillwater:
My company has need of the new "Sign in with Apple" functionality.

https://developer.apple.com/sign-in-with-apple/

I'll give it a stab and submit a PR at some point soon.

juanscolari:
@tylerb do you have any WIP/Draft for this feature to show us?

tylerb:
Not yet. I'm hoping to get something up this week.

guilhermehubner:
@tylerb any update on it?

tylerb:
Unfortunately not. Things have been crazy at work and I've not had a chance to get to this. Still fully intend to, though. One of our other developers may jump on it.

wunderkind2k1:
Isn't this done?

wunderkind2k1:
I think this issue can be closed - what do you guys think?

yaronius:
@markbates @bentranter please close

PR #293, opened 2019-10-08
Sign In with Apple

lucjross:
https://github.com/markbates/goth/issues/290


bentranter:
Thanks @lucjross! Pretty sure the test failure wasn't your fault, so I've restarted the job that failed. I'll merge once that passes.

providers/apple/apple.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func (Provider) UnmarshalSession(data string) (goth.Session, error) {
+	s := &Session{}
+	err := json.NewDecoder(strings.NewReader(data)).Decode(s)
+	return s, err
+}
+
+func (Provider) FetchUser(goth.Session) (goth.User, error) {
+	return goth.User{}, errors.New("not implemented")
```
lucjross: I couldn't find a user profile endpoint for Apple.

lucjross:
Please, don't merge yet. We want to try and get FetchUser implemented.

bentranter:
Sure, no problem @lucjross, I'll wait.

lucjross:
@bentranter, I still haven't found a user profile endpoint for Apple, but I made FetchUser a little better by using the unique ID from the ID token on the access token response.

Looks like you're trying to support Go versions down to 1.7. If so I'll need to pull out that `jwx` library.

bentranter:
@lucjross I wouldn't worry about supporting all the way back to 1.7, since they're all way past EOL – I probably should've dropped support for them a long time ago. If you want, you can remove the failing old versions from the test matrix, if not, I can do it and you can rebase your PR.

lucjross:
Cool, thanks for the flexibility on that. I think we've implemented everything needed on our end.

PR #294, opened 2019-10-09
Feature - Slack Provider - Progressive Profile Fetching

Rican7:
# Scope

This PR addresses some usability issues with the current Slack provider, some security concerns, some HTTP client response closing issues, code formatting and linting issues, and greatly improves unit-test code coverage **for the Slack provider only**.

This was developed in a TDD fashion and the code has already been tested in an integration via our downstream fork.

### Most of this PR (in terms of lines of code) is in fact tests.
Don't let the size scare you. 😅

## Motivations

While working on a project that uses the Goth library, and more specifically the Slack provider, I ran into an issue where this library was essentially _forcing_ the use of [the `users:read` scope](https://api.slack.com/scopes/users:read), due to the way that it fetches the user's information at [the `users.info` endpoint](https://api.slack.com/methods/users.info).

While the Slack provider allows for a custom set of "scopes" to be defined via the last parameter of `slack.New`, the provider's `FetchUser` method **always** hits [the `users.info` endpoint](https://api.slack.com/methods/users.info) that **requires** [the `users:read` scope](https://api.slack.com/scopes/users:read). This unfortunately causes a "silent error" of sorts, as the response code is never even checked and the data won't be returned without that scope in place.

Ideally, and as this PR implements, the `FetchUser` method would return the "basic" profile (identity) of the user that it can get with [the `auth.test` endpoint](https://api.slack.com/methods/auth.test) that it must hit anyway, and then only attempt to grab the additional information if the required scope is present.

## Specifics

 - Adds comments and fixing formatting, to appease golint
 - Defines the `users:read` Slack provider scope as a constant
 - Greatly increases test coverage for the Slack provider
 - Refactors `providers/slack.Provider.FetchUser` to use the initially fetched data as a "basic" profile and only fetch the other, more "expansive", user-data if the proper scope has been included in the provider
     - This improves the Slack provider to only perform an operation when its able to (has the scope) and allows for users of the provider to choose whether they want to use the `users:read` scope, which is a scope that Slack considers to be advanced enough to "warn" about in Slack apps.
     - Also, this enables the Slack provider to be used with the newer "Sign in with Slack" flow and potentially ask for more scopes/permission "progressively".
 - Adds a missing HTTP client response body closure and removing redundant and unnecessary closures
 - Adds proper handling and testing for bad Slack user-info HTTP responses

### DIFF
--- a/providers/slack/slack.go
+++ b/providers/slack/slack.go
@@ -11,10 +11,12 @@ import (
 	"net/url"
 
 	"fmt"
+
 	"github.com/markbates/goth"
 	"golang.org/x/oauth2"
 )
 
+// URLs and endpoints
 const (
 	authURL         string = "https://slack.com/oauth/authorize"
 	tokenURL        string = "https://slack.com/api/oauth.access"
@@ -56,6 +58,7 @@ func (p *Provider) SetName(name string) {
 	p.providerName = name
 }
 
+// Client returns the http.Client used in the provider.
 func (p *Provider) Client() *http.Client {
 	return goth.HTTPClientWithFallBack(p.HTTPClient)
 }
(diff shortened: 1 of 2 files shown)


Rican7:
> Looks good – thanks for all your hard work @Rican7! 🎉

Thank you! Thanks for the quick response, merge, and tagging! 😀

PR #295, opened 2019-10-24
apple -- make SecretParams properties public

lucjross:
A slight correction to my last contribution.


lucjross:
Thanks again @bentranter !

PR #301, opened 2019-12-16
Export a non-collidable context key for setting provider in context

### LINKED ISSUE #238: gothic: Attempt to get Provider Name from go context probably shouldn't use a string key
Right now, the final attempt to determine the provider name is as follows, in gothic.go:

```
//  try to get it from the go-context's value of "provider" key
if p, ok := req.Context().Value("provider").(string); ok {
	return p, nil
}
```

In trying to get gothic to play nicely with labstack/echo, I wanted to use the go-context to pass along the provider name-- basically transferring c.Param("provider") into the context.  However, attempts to do so result in go-lint errors (_Should not use basic type string as key in context.WithValue_); you aren't supposed to use plain strings as go-context keys.

One solution would be for Gothic to define a new key type, such as
```
type ProviderContextKey struct{}
```
And then use that for (or in addition to) this check.  Then one could simply stuff the provider in using this more idiomatic key type.  I tried this myself using a custom override of the GetProviderName and it seems to work fine.  In my handler, I wound up with:

```
nr := c.Request().WithContext(context.WithValue(c.Request().Context(), providerContextKey{}, c.Param("provider")))
c.SetRequest(nr)
```

Which is a mouthful, but seems to work fine.   Is there a better way?  There's a goth/echo fork (https://github.com/MilyMilo/goth) but I prefer not to stray from mainline.

zinefer:
This pull request add an exported [key constant](https://blog.golang.org/context#TOC_3.2.) that can be used when adding the `provider` to a context for `gothic`. Using this exported key prevents the go-lint error `Should not use basic type string as key in context.WithValue`.

```golang
provider := "github"
ctx := context.WithValue(req.Context(), gothic.ProviderParamKey, provider)
req = req.WithContext(ctx)
```

Closes #238

### DIFF
--- a/gothic/gothic.go
+++ b/gothic/gothic.go
@@ -35,6 +35,10 @@ var defaultStore sessions.Store
 
 var keySet = false
 
+type key int
+// ProviderParamKey can be used as a key in context when passing in a provider
+const ProviderParamKey key = iota
+
 func init() {
 	key := []byte(os.Getenv("SESSION_SECRET"))
 	keySet = len(key) != 0
@@ -262,6 +266,11 @@ func getProviderName(req *http.Request) (string, error) {
 
 	//  try to get it from the go-context's value of "provider" key
 	if p, ok := req.Context().Value("provider").(string); ok {
+		return p, nil
+	}
+
+	// try to get it from the go-context's value of providerContextKey key
+	if p, ok := req.Context().Value(ProviderParamKey).(string); ok {
 		return p, nil
 	}
 


zinefer:
So I was just wondering if maybe exporting a method that set this parameter in the context for us. Maybe something like this in the `gothic.go`?

```bash
func GetContextWithProvider(req *http.Request, provider string) *http.Request {
	return req.WithContext(context.WithValue(req.Context(), ProviderParamKey, provider))
}
```

danielbprice:
(As the person who filed #238) this looks good to me.  Thanks!

PR #315, opened 2020-02-10
Added custom auth code options for google

gregadams4:
Might just be missing something, but couldn't find a way currently to set access_type=offline to get a refresh token for this provider

### DIFF
--- a/providers/google/google.go
+++ b/providers/google/google.go
@@ -32,13 +32,13 @@ func New(clientKey, secret, callbackURL string, scopes ...string) *Provider {
 
 // Provider is the implementation of `goth.Provider` for accessing Google.
 type Provider struct {
-	ClientKey    string
-	Secret       string
-	CallbackURL  string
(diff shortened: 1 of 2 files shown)


bentranter:
@gregadams4 Thanks for bringing this up! I we should instead set `access_type=offline` by default. As stated in Google's documentation,
> Set the value to `offline` if your application needs to refresh access tokens when the user is not present at the browser.

I feel like that covers the authentication use case for Goth. We also used to set it by default on the old `gplus` provider, so I think it'd be better to have that as the default here for consistency.

jmburges:
🤞 this gets merged soon!

uded:
Can someone check, fix and merge this? This is a crucial update from the perspective of using Google auth for something more than just login...

uded:
After using a newly merged #396 I have to say this would be a bit better. With a default for `access_type=offline`. But sometimes I do need to add `refresh_token` or something like that...

PR #317, opened 2020-02-10
Make decodeJWT use URLEncoding without padding

### LINKED ISSUE #316: decodeJWT should decode base64url
Since JSON Web Tokens (afaik) are encoded using base64url by default, the decoding method should reflect this. 

https://github.com/markbates/goth/blob/8a9b88125edf10f2441ec2cc9aa6f40c39e50186/providers/openidConnect/openidConnect.go#L398-L402

Could instead be simply 

```
decodedPayload, err := base64.URLEncoding.WithPadding(base64.NoPadding).DecodeString(encodedPayload)
```
since base64url has no padding.

Alternatively, a check to correct base64url-encoded tokens that translates them to a padded standard encoded string could be introduced, whichever is preferred?

oscarlofwenhamn:
Fixes #316

### DIFF
--- a/providers/openidConnect/openidConnect.go
+++ b/providers/openidConnect/openidConnect.go
@@ -393,13 +393,8 @@ func decodeJWT(jwt string) (map[string]interface{}, error) {
 		return nil, errors.New("jws: invalid token received, not all parts available")
 	}
 
-	// Re-pad, if needed
-	encodedPayload := jwtParts[1]
-	if l := len(encodedPayload) % 4; l != 0 {
-		encodedPayload += strings.Repeat("=", 4-l)
-	}
+	decodedPayload, err := base64.URLEncoding.WithPadding(base64.NoPadding).DecodeString(jwtParts[1])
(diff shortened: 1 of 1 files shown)


oscarlofwenhamn:
A review would be very helpful as this is a major bug in our end. @willemvd, @bentranter, or anyone else?
I see @techknowlogick has been active, but can not find them in the Reviewers list for assigning.

oscarlofwenhamn:
Cheers, thanks for the review and merge!

techknowlogick:
@oscarlofwenhamn I'm not a maintainer of goth, I am one of the leads of Gitea and so we use goth a lot so I happen poke at goth a lot. Although @bentranter I'd be happy to volunteer some of my time to assist with maintenance of goth.

oscarlofwenhamn:
@techknowlogick That makes a lot of sense, and explains why I've seen you as much in both places. Hope it wasn't a bother getting tagged into the thread!

PR #318, opened 2020-02-11
Add login.gov provider

andrewzah:
This PR adds support for [login.gov](https://login.gov) which uses a [customized OIDC flow](https://developers.login.gov/oidc/).

I adapted the oidc provider and tried to adhere to the style of the project; please let me know if you have any feedback.


providers/logingov/logingov.go
```diff
@@ -0,0 +1,434 @@
...
+)
+
+// These vars are necessary for login.gov auth to work.
+// These default to the login.gov sandbox values
+//   * See the getEnv fn below.
+// AcrValues: Only accepts LOA1 or LAO3 urls.
+//   * See https://developers.login.gov/oidc/#authorization
+//   * Currently only /loa/1 and /ial/1 work, as per the gsa-tts slack.
+var (
+	acrValues    = getEnv("LOGINGOV_ACR_VALUES", "http://idmanagement.gov/ns/assurance/loa/1")
```
techknowlogick: I recommend avoiding getEnv in the library in case the application using this library needs to have multiple instances of this login provider (possibly an app has an area where they only need LOA1 capabilities, but then a section of it needs upgraded LOA3 capabilities). What you could instead do is hardcode your default of LOA1 here, and then have a setup function (see: https://github.com/markbates/goth/pull/276/files#diff-a86775ae2ccd437b48a63db0a7239c49R52 for an example where Gitea login provider needs different setup than provided in default interface) where acr and approved emails are passed in via argument, and save into the provider struct.

azah: Duly noted. I updated it to have a NewCustomisedURL() fn, and now that fn sets AcrValues / ApprovedEmails.

azah:
@techknowlogick I used this branch ([b2f29be](https://github.com/azah/goth/tree/login_gov-provider)) in a fork of gitea (diff: [4484323](https://github.com/azah/gitea/commit/4484323bdef7aa57e38b07eeb9c44f904ff833f5)), and I ran into an issue. I was able to sign in successfully with login.gov to gitea and create a new account, but when someone else signed in with login.gov, they logged into my gitea account. Do you think this is an issue with my goth implementation? Or with my gitea one?

techknowlogick:
@azah in terms of changes you made to gitea it looks like the code changes there are correct. Could you test with a different provider (github is easiest one to test with as we have many implementations of gitea using it, so we know that integration is solid) to see if same thing still happens? If same thing still happens with a different provider then it is likely due to a configuration of gitea, otherwise then it is likely due to something in this PR.

azah:
@techknowlogick good idea. With github auth it worked correctly, so I'll take a look at this PR again on monday to see what might be causing the issue.

azah:
@techknowlogick as far as I'm aware, each call gets a new `Session`, so I don't understand why different users would map to the same gitea account. I ran the example app in two firefox windows side by side with different sandbox-login.gov accounts at the same time, and had no issues.

Could it be something to do with creating a new account in gitea?

techknowlogick:
I've just created a login.gov account (I don't work for USG so I didn't realize it was open to any random person including non-americans), so I can help tetst. I haven't figured out how to make an oauth2 application so I can test with a local gitea install, do you have any insight into how to do that?

azah:
@techknowlogick there's a process to getting approved. If you send a message to zah@andrewzah.com (or reply with your email/twitter) I can send you an issuer id/redirect_uri for testing goth and/or gitea.

techknowlogick:
sent. ty :)

azah:
@techknowlogick thanks, I sent you an email. I'm still using [commit 4484323](https://github.com/azah/gitea/commit/4484323bdef7aa57e38b07eeb9c44f904ff833f5) from [my gitea fork](https://github.com/azah/gitea/tree/logingov-provider)

PR #321, opened 2020-03-05
Add support for multiple Apple public keys

dstapleton92:
This Sign In With Apple (SIWA) implementation has been failing intermittently. After some investigation, I have determined the cause of the intermittent failure. Before, Apple only provided a single public key in the array hosted at this url: https://appleid.apple.com/auth/keys.

The original author of this apple provider assumed (wish Apple had better docs!) there would only ever be one key, so always used `keys[0]`. Now, Apple has added 2 more keys (making 3 total). The correct procedure is to retrieve the key id (`kid`) from the header of the Apple JWT and then use the corresponding key from the `keys` array to verify the JWT.

I also removed a comment mentioning the need to fetch the keys from Apple less frequently by respecting the cache headers. The headers Apple sets on that response indicate that it should not be cached.

### DIFF
--- a/providers/apple/session.go
+++ b/providers/apple/session.go
@@ -75,6 +75,7 @@ func (s *Session) Authorize(provider goth.Provider, params goth.Params) (string,
 
 	if idToken := token.Extra("id_token"); idToken != nil {
 		idToken, err := jwt.ParseWithClaims(idToken.(string), &IDTokenClaims{}, func(t *jwt.Token) (interface{}, error) {
+			kid := t.Header["kid"].(string)
 			claims := t.Claims.(*IDTokenClaims)
 			vErr := new(jwt.ValidationError)
 			if !claims.VerifyAudience(p.clientId, true) {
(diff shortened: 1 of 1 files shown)


dstapleton92:
@mca312 In case you've been pulling your hair out over this like me...

providers/apple/session.go
```diff
@@ -100,12 +101,18 @@ func (s *Session) Authorize(provider goth.Provider, params goth.Params) (string,
 			}
 
 			// get the public key for verifying the identity token signature
-			// todo: respect Cache-Control header and retrieve this less frequently
-			set, err := jwk.FetchHTTP(idTokenVerificationKeyEndpoint, jwk.WithHTTPClient(p.httpClient))
+			set, err := jwk.FetchHTTP(idTokenVerificationKeyEndpoint, jwk.WithHTTPClient(p.Client()))
```
dstapleton92: Switching this to use `p.Client()` so that the fallback client will be retrieved if one was not set during Apple controller initialization.

bentranter:
This is out in https://github.com/markbates/goth/releases/tag/v1.61.3.

mca312:
@dstapleton92 Thanks.  I did something similar when verifying the JWT.  But I'm doing this right before the session.Authorize

```
kid := t.Header["kid"]
jwk := set.LookupKeyID(kid.(string))
pubKeyIface, _ := jwk[0].Materialize()
```

dstapleton92:
Dang. Didn't know about `set.LookupKeyID(...)`. Oh well. What I did is pretty similar to their implementation.

PR #328, opened 2020-04-06
Added Kakao Talk provider

kenkoii:
This PR adds Kakao as a new provider for this package


kenkoii:
Would it be possible to have this reviewed and perhaps merged? will be happy to work on it if there are issues :bow:

PR #332, opened 2020-05-02
Ensure response body is closed in lastfm provider

bvwells:


bvwells:
Thanks for the review and merge @bentranter!

Issue #334, opened 2020-05-02
Go report card


bvwells:
I noticed a couple of minor typos (#333) and was wondering whether it is worth adding go report card as I suspect that this would have highlighted these minor issues. 

More than happy to open a PR if you want it added.

P.S. Nice repo, it's a really helpful resource!

bvwells:
Added a PR, but more than happy to abandon if this is not needed.

bentranter:
I think it's useful, especially for me to track any changes over time. Thanks for that PR!

PR #335, opened 2020-05-02
Add go report card to README

bvwells:
Add go report card to README to address issue #334. Hope this is useful.


bvwells:
No problems. Thanks for the review and merge!

PR #341, opened 2020-05-20
Added Atlassian Provider

lettfeti:
Not exactly sure what to say here, I've been using Goth and I wanted to use it with their 3LO for apps flow


lettfeti:
removed the binary, and rebased onto a single commit for cleanliness

lettfeti:
updated pr to resolve conflicts with okta addition, anyone want to take a second look?

PR #344, opened 2020-06-03
Added "profile" in array of scopes in file google.go.

obsessedwithperfection:
email and profile are default scopes of google api oauth2.0.

When a developer does not provide scope values then by default it can now fetch both email and basic profile information from google api oauth2.0.

Very handy for beginners.

### DIFF
--- a/providers/google/google.go
+++ b/providers/google/google.go
@@ -32,13 +32,13 @@ func New(clientKey, secret, callbackURL string, scopes ...string) *Provider {
 
 // Provider is the implementation of `goth.Provider` for accessing Google.
 type Provider struct {
-	ClientKey       string
-	Secret          string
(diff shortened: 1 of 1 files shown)


bentranter:
Thanks @obsessedwithperfection, but I have one concern with this change. For folks who already use the Google provider, I'm worried this change will cause any user signing in to have to re-authenticate because of the scope change. Do you know if that's the case, given that email and profile are the default scopes?

techknowlogick:
Thank you for this PR. I have the same concerns as Ben, as through testing this I can confirm that re-auth would be required. As such this would be "breaking" in my opinion, and so I think I will close this PR now, however thinking about longer term approaches perhaps we could have a new section of documentation on different ways to use each provider outside of defaults.

PR #356, opened 2020-08-26
Minor adjustment in GetState when apple posts callback (instead of get). Fixes #360

### LINKED ISSUE #360: State validation when using "sign in with apple" (sometimes) broken
# Description
If I try to sign in with apple and the scopes "name" and "email", the login is not working.

# Analysis/ Fix
Apple uses a POST callback, when we use "name" or "email" scope. This results in the state encoded in the form parameter instead of url. The current implementation fo GetState does not change its behaviour if the callback is sent with POST instead of GET. This results in an error in the state comparison and a broken login.

wunderkind2k1:
Fixes #360

I couldn't get sign in with apple working with "email" and "name" scopes. The callback in this case is send as a "special" POST request which encodes the state into the request Form and not in the URL.

This led to errors on state comparison on CSFR mitigation. The solution is to fetch the state from the form in case of an apple post request.

We just put the existing code from the already merged #339 into the addtional destination in GetState.


gothic/gothic.go
```diff
@@ -99,6 +99,14 @@ var SetState = func(req *http.Request) string {
 // This is used to prevent CSRF attacks, see
 // http://tools.ietf.org/html/rfc6749#section-10.12
 var GetState = func(req *http.Request) string {
+
+	params := req.URL.Query()
+	if params.Encode() == "" && req.Method == "POST" {
+		req.ParseForm()
```
macuenca, line 105: Hey @wunderkind2k1, I just realized we actually stumbled upon the same issue. I opened https://github.com/markbates/goth/pull/357 with my proposal for the fix, but yours is more appropriate as it handles it in the `GetState()` function.

May I suggest to avoid calling `ParseForm()` here and use `req.FormValues("state")`? FormValues calls ParseForm internally, so there's not much difference, but the code will be shorter. What about doing something like this instead?

```go
var GetState = func(req *http.Request) string {
	if req.Form.Encode() == "" && req.Method == http.MethodPost {
                return req.FormValues("state")
	}

	return req.URL.Query().Get("state")
}
```

wunderkind2k1: Good idea. Thank you. 9a337777e3823df0d799d3013f8b9b163dc41480 should do it.

wunderkind2k1: Hey @macuenca - I have made your suggested changes. Maybe you are so kind and can close your pull request #357 then please?

gothic/gothic.go
```diff
@@ -99,6 +99,10 @@ var SetState = func(req *http.Request) string {
 // This is used to prevent CSRF attacks, see
 // http://tools.ietf.org/html/rfc6749#section-10.12
 var GetState = func(req *http.Request) string {
+	params := req.URL.Query()
+	if params.Encode() == "" && req.Method == "POST" {
```
macuenca, line 103: Consider using `http.MethodPost` here instead of hardcoding "POST".

gothic/gothic_test.go
```diff
@@ -251,6 +251,16 @@ func Test_StateValidation(t *testing.T) {
 	a.Error(err)
 }
 
+func Test_AppleStateValidation(t *testing.T) {
+	a := assert.New(t)
+	appleStateValue := "xyz123-#"
+	form := url.Values{}
+	form.Add("state", appleStateValue)
+	req, _ := http.NewRequest("POST", "/", strings.NewReader(form.Encode()))
```
macuenca, line 259: Consider using `http.MethodPost` here instead of hardcoding "POST".

gothic/gothic.go
wunderkind2k1: https://github.com/markbates/goth/pull/356/commits/1ace300f3fc29e2458b7d7ec3ead4d33efa86968 should do it. Thanks for pointing me to it @macuenca

gothic/gothic.go
wunderkind2k1: Yes. See https://github.com/markbates/goth/pull/356/commits/1ace300f3fc29e2458b7d7ec3ead4d33efa86968. Thank you.

wunderkind2k1:
Today I rebased the maste branch - hey guys. What about this PR - is  anything missing? Tests included...

wunderkind2k1:
@bentranter sorry to mention you here directly but I think this PR might resolve more than one issue. Do you think its possible to have a look at it?

bentranter:
Hey @wunderkind2k1, the change looks great, but the diff and changeset is coming out a bit weird. Could you try rebasing your PR one more time?

go.mod
```diff
@@ -1,14 +1,20 @@
-module github.com/markbates/goth
+module github.com/damonkeys/goth
```
techknowlogick, line 1: Please revert this line

go.mod
```diff
@@ -1,14 +1,20 @@
...
+	cloud.google.com/go v0.63.0 // indirect
+	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/dgrijalva/jwt-go v3.2.0+incompatible
 	github.com/gorilla/mux v1.6.2
 	github.com/gorilla/pat v0.0.0-20180118222023-199c85a7f6d1
 	github.com/gorilla/sessions v1.1.1
 	github.com/jarcoal/httpmock v0.0.0-20180424175123-9c70cfe4a1da
 	github.com/lestrrat-go/jwx v0.9.0
 	github.com/markbates/going v1.0.0
+	github.com/markbates/goth v1.65.0
```
techknowlogick, line 15: This is probably an artifact due to line above

go.mod
wunderkind2k1: Oops. Sorry. Should be fixed with 59a4506cdec732fef480e4bac5a34a3550836cdb

go.mod
wunderkind2k1: should be resolved due to 59a4506cdec732fef480e4bac5a34a3550836cdb

wunderkind2k1:
> Hey @wunderkind2k1, the change looks great, but the diff and changeset is coming out a bit weird. Could you try rebasing your PR one more time?

@bentranter is it better now?

techknowlogick:
There diff/changeset still are a bit off, you can see here https://github.com/markbates/goth/pull/356/files what we see, and it contains things that have already been merged into the master branch.

wunderkind2k1:
I am unsure what to do. @techknowlogick could you please be a little bit more specific?

yaronius:
@techknowlogick @bentranter I re-applied changes from this PR in https://github.com/markbates/goth/pull/395, the diff seems OK now. But some windows tests got broken 😢 I think many people need those changes, so I'm ready to help with fixing the tests

PR #357, opened 2020-08-31
Capture state when coming from a POST request

macuenca:
Sign-in with Apple sends the `code` and `state` values after the authentication exchange has been successful using a POST request. The current code is not expecting it and returns a 405 response with the message `state token mismatch`.

This fixed it for me, so I'm putting it under consideration to be merged in the code. If this is the wrong approach, I would appreciate any input you can provide so I can make the proper changes in my code.


macuenca:
This is same is better addressed by @wunderkind2k1 in #356

PR #386, opened 2021-01-21
Export OpenIDConfig in openIdConnect provider

subhan-nadeem:
This PR exports the `OpenIDConfig` field in the `openIdConnect` provider. The user may want to roll their logic when attempting to utilize the endpoints defined in `OpenIDConfig` such as the `AuthEndpoint` or the `RefreshEndpoint`.

In my case, the internal `RefreshToken` call isn't sufficient; it utilizes the external `oauth` library's `RefreshToken` call that unmarshalls the refresh token response into a struct containing `AccessToken` and `RefreshToken`, discarding the refreshed `id_token` in the response body (part of the [OpenID Connect spec](url)). Since `id_token` isn't a part of the core oauth spec, either the `RefreshToken` call needs to roll its own request, or alternatively, export the configuration that provides the refresh endpoint which allows the user to utilize it as they wish.

### DIFF
--- a/providers/openidConnect/openidConnect.go
+++ b/providers/openidConnect/openidConnect.go
@@ -54,8 +54,8 @@ type Provider struct {
 	Secret       string
 	CallbackURL  string
 	HTTPClient   *http.Client
+	OpenIDConfig *OpenIDConfig
 	config       *oauth2.Config
-	openIDConfig *OpenIDConfig
 	providerName string
 
 	UserIdClaims    []string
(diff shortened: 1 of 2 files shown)


bentranter (approved):
Looks good to me, thanks @subhan-nadeem!

bentranter:
Released in [v1.67.0](https://github.com/markbates/goth/releases/tag/v1.67.0).

PR #387, opened 2021-01-22
Support User.rawData in Salesforce provider

mxaly:
This PR adds support for User.rawData in salesforce provider. 

Since there are cases when SF custom data, not supported out of the box by goth is needed, `User.rawData` should be populated with raw result of `FetchUser` request. 
In my case `urls.custom_domain` is needed to determinate whats the host of salesforce user. Storing it `User.rawData` will do the job :)

### DIFF
--- a/providers/salesforce/salesforce.go
+++ b/providers/salesforce/salesforce.go
@@ -3,6 +3,7 @@
 package salesforce
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


bentranter (approved):
Looks good to me, thanks @mxaly!

bentranter:
Released in [v1.67.0](https://github.com/markbates/goth/releases/tag/v1.67.0).

PR #391, opened 2021-01-25
Add a Refresh Token function to the OpenID Connect provider that also returns the refreshed OpenID id_token

subhan-nadeem:
[The ID token is a fundamental part of the OpenID connect refresh token flow](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) (`id_token`). but is not part of the OAuth flow.

The existing `RefreshToken` function in the OpenID Connect provider leverages the OAuth library's refresh token mechanism, ignoring the optionally refreshed `id_token`. As a result, a new function needs to be exposed (rather than changing the existing function, for backwards compatibility purposes) that also returns the `id_token` in the OpenID refresh token flow API response. 

This PR creates and exposes such a flow with the function `RefreshTokenWithIDToken`. I've written this logic internally for our project at Tesla because we cannot use the `RefreshToken` endpoint due to it discarding the `id_token` response field, but figured it should be a part of the original `goth` library since this logic conforms to the Open ID connect spec.

### DIFF
--- a/providers/openidConnect/openidConnect.go
+++ b/providers/openidConnect/openidConnect.go
@@ -81,12 +81,11 @@ type OpenIDConfig struct {
 type RefreshTokenResponse struct {
 	AccessToken string `json:"access_token"`
 
-
 	// The OpenID spec defines the ID token as an optional response field in the
 	// refresh token flow. As a result, a new ID token may not be returned in a successful
 	// response.
 	// See more: https://openid.net/specs/openid-connect-core-1_0.html#RefreshingAccessToken
-	IdToken     string `json:"id_token, omitempty"`
+	IdToken string `json:"id_token, omitempty"`
 
 	// The OAuth spec defines the refresh token as an optional response field in the
 	// refresh token flow. As a result, a new refresh token may not be returned in a successful
--- a/providers/openidConnect/openidConnect.go
+++ b/providers/openidConnect/openidConnect.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


providers/openidConnect/openidConnect.go
```diff
@@ -197,6 +214,46 @@ func (p *Provider) RefreshToken(refreshToken string) (*oauth2.Token, error) {
...
+
+	resp, err := p.Client().Do(req)
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("Non-200 response from RefreshToken: %d, WWW-Authenticate=%s", resp.StatusCode, resp.Header.Get("WWW-Authenticate"))
+	}
+
+	body, err := ioutil.ReadAll(resp.Body)
```
bentranter, line 244: The error from `ioutil.ReadAll` should be checked and returned,
```suggestion
	body, err := ioutil.ReadAll(resp.Body)
	if err != nil {
	    return nil, err
	}
```

bentranter (changes requested):
One tiny change, after that, LGTM!

providers/openidConnect/openidConnect.go
subhan-nadeem: Addressed! Thank you

subhan-nadeem:
Change addressed @bentranter , thank you

bentranter (approved):
LGTM, thanks again @subhan-nadeem!

mrkschan:
Is there any specific reason not to parse `expires_in` of the refresh token response as per https://openid.net/specs/openid-connect-core-1_0.html#RefreshTokenResponse?

Without such field, caller of `RefreshTokenWithIDToken()` has to parse the JWT again just to obtain the expiry.

PR #397, opened 2021-02-19
Add NewCustomisedURL function to the OpenID connect provider

norbertfenk-striva:
Currently, there was no function to use the OpenID connect provider without auto-discovery, but I think this one is a possible use case for others too.


SzilardS:
@markbates @bentranter this seems a great improvement and much needed for us also. Could you please merge it? (this would prevent us to maintain a forked repo)

PR #405, opened 2021-06-16
Make it possible to set bot_prompt query param for line

k-yomo:
To display the option to add your LINE Official Account as a friend, we need to pass `bot_prompt` query parameter for the auth url.
This PR adds `SetBotPrompt` function to set the query param.

details for the `bot_promt` is written in the document below.
https://developers.line.biz/en/docs/line-login/link-a-bot/#redirect-users

bot-prompt 1dd4e585

### DIFF
--- a/providers/line/line_test.go
+++ b/providers/line/line_test.go
@@ -57,7 +57,7 @@ func Test_SetBotPrompt(t *testing.T) {
 	session, err := p.BeginAuth("test_state")
 	s := session.(*line.Session)
 	a.NoError(err)
-	a.Contains(s.AuthURL, "&bot_prompt=normal")
+	a.Contains(s.AuthURL, "bot_prompt=normal")
 }
 
 func provider() *line.Provider {
--- a/providers/line/line_test.go
+++ b/providers/line/line_test.go
@@ -48,6 +48,18 @@ func Test_SessionFromJSON(t *testing.T) {
 	a.Equal(s.AccessToken, "1234567890")
 }
 
+func Test_SetBotPrompt(t *testing.T) {
+	t.Parallel()
+	a := assert.New(t)
+
+	p := provider()
+	p.SetBotPrompt("normal")
+	session, err := p.BeginAuth("test_state")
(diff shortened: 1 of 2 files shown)


techknowlogick (approved):

bentranter (approved):
LGTM! (pending CI)

providers/line/line_test.go
```diff
@@ -48,6 +48,18 @@ func Test_SessionFromJSON(t *testing.T) {
...
+func Test_SetBotPrompt(t *testing.T) {
+	t.Parallel()
+	a := assert.New(t)
+
+	p := provider()
+	p.SetBotPrompt("normal")
+	session, err := p.BeginAuth("test_state")
+	s := session.(*line.Session)
+	a.NoError(err)
+	a.Contains(s.AuthURL, "&bot_prompt=normal")
```
bentranter, line 60: Looks like this bit of the test can be flaky depending on the order of the query params,

```
 --- FAIL: Test_SetBotPrompt (0.00s)
    line_test.go:60: 
        	Error Trace:	line_test.go:60
        	Error:      	"https://access.line.me/oauth2/v2.1/authorize?bot_prompt=normal&client_id=&redirect_uri=%2Ffoo&response_type=code&state=test_state" does not contain "&bot_prompt=normal"
        	Test:       	Test_SetBotPrompt
```

for example, it may create the URL param prefixed with `?` or `&`. It might be better not to check that leading character, and just do,

```go
a.Contains(s.AuthURL, "bot_prompt=normal")
```

bentranter (changes requested):
One comment about the tests, after that it should be good to merge. Thanks for the contribution @k-yomo!

providers/line/line_test.go
k-yomo: Thanks for the good catch! fixed to do `a.Contains(s.AuthURL, "bot_prompt=normal")`!

bentranter (approved):
LGTM, thanks @k-yomo!

PR #409, opened 2021-07-03
Add Wizards of the Coast provider

huntrar:
Wizards of the Coast / Hasbro is a game publisher, notably of Dungeons & Dragons and Magic: The Gathering. An account at myaccounts.wizards.com may be used to login via this provider.

### DIFF
--- a/providers/wizards/wizards.go
+++ b/providers/wizards/wizards.go
@@ -1,0 +1,188 @@
+// Package wizards implements the OAuth2 protocol for authenticating users through Wizards of the Coast "Platform".
+// This package can be used as a reference implementation of an OAuth2 provider for Wizards.
+package wizards
+
+import (
+	"encoding/json"
+	"fmt"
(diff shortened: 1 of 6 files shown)


providers/wizards/wizards.go
```diff
@@ -0,0 +1,189 @@
+// Package wizards implements the OAuth2 protocol for authenticating users through Wizards of the Coast "Platform".
+// This package can be used as a reference implementation of an OAuth2 provider for Wizards.
+package wizards
+
+import (
+	"encoding/json"
+	"io"
+	"net/http"
+
```
techknowlogick, line 9: tiny nit, but fmt should probably be grouped together with the imports above.

techknowlogick (commented):
I'm not a maintainer of this library, however just one nit that may be of benefit mentioned below. You may also wish to add this provider to the example and to docs as well.

PR #417, opened 2021-08-04
Add IDToken for google provider

wgjak47:
like https://github.com/markbates/goth/pull/297, add IDToken field for google provider


bentranter (approved):
LGTM, thanks @wgjak47!

PR #420, opened 2021-08-20
WIP: upgrade jwx/jwk

hjanuschka:
hey 🤚 

many thx for the great lib, sadly i need to use a bit more updated version of jwx/jwk.
this PR will try to upgrade it.

consider it WIP - just need to see if test's pass

regards


go.mod
```diff
@@ -10,11 +10,9 @@ require (
 	github.com/gorilla/pat v0.0.0-20180118222023-199c85a7f6d1
 	github.com/gorilla/sessions v1.1.1
 	github.com/jarcoal/httpmock v0.0.0-20180424175123-9c70cfe4a1da
-	github.com/lestrrat-go/jwx v0.9.0
+	github.com/lestrrat-go/jwx v1.2.5
```
u-nation, line 13: If possible, could you please change the version to the latest v1.2.15?
https://github.com/lestrrat-go/jwx/releases

It's a patch version change, so I don't think it will have any bad effects:)

PR #424, opened 2021-09-13
Add SetAccessType for Google OAuth

yyewolf:
Useful to request the Google OAuth API and then using other endpoints (admin sdk / drive etc...)


bentranter (approved):
Looks good, thanks @yyewolf!

PR #447, opened 2022-02-22
bitbucket: check the is_primary and is_confirmed properties on the email address

### LINKED ISSUE #446: Bitbucket email confirmed/primary properties are not checked
Looking at my browser's requests when I log in to https://bitbucket.org :

```
GET https://bitbucket.org/!api/2.0/user/emails
```

```json
{"pagelen":10,"values":[{"is_primary":true,"links":{"self":{"href":"https:\/\/bitbucket.org\/!api\/2.0\/user\/emails\/username@example.org"}},"type":"email","email":"username@example.org","is_confirmed":true}],"page":1,"size":1}
```

The `is_confirmed` and `is_primary` properties are not being checked by goth:
https://github.com/markbates/goth/blob/19c3059c9ed44995940c8222a8eda10f6b604e01/providers/bitbucket/bitbucket.go#L158-L171

I believe goth should be checking those properties from Bitbucket and refusing to return an unconfirmed email address, similarly to what it does for github:
https://github.com/markbates/goth/blob/19c3059c9ed44995940c8222a8eda10f6b604e01/providers/github/github.go#L204-L209

Unfortunately Bitbucket doesn't seem to document the emails response schema anywhere, the best we have is:
https://developer.atlassian.com/cloud/bitbucket/rest/api-group-users/#api-user-emails-get
which says:
> Returns all the authenticated user's email addresses. Both confirmed and unconfirmed.

I can make a PR, if you don't object.

pjcdawkins:
Fixes #446 - see the issue for more detail

This PR includes a little bit of refactoring, because passing the `access_token` as a URL parameter seems like an older thing. [Bitbucket's API documentation](https://developer.atlassian.com/cloud/bitbucket/rest/api-group-users/#api-user-emails-get) suggests passing it as a Bearer token.

### DIFF
--- a/providers/bitbucket/bitbucket.go
+++ b/providers/bitbucket/bitbucket.go
@@ -5,10 +5,8 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"net/http"
-	"net/url"
 
 	"github.com/markbates/goth"
 	"golang.org/x/oauth2"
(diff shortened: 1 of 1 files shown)


techknowlogick (approved):
Thanks for this PR! It looks great, one thing I am concerned with is that now if there is no email (primary/confirmed or otherwise), the login will error out. However I am approving this as technically emails are always fetched previously, so this is a case that should already be covered. (I'm just explicitly stating this case as acknowledge so that in the future someone can point to my comment and see that this is the expected behaviour)

pjcdawkins:
That was quick - thanks @techknowlogick

PR #451, opened 2022-03-26
Use Bearer token in Slack

Steve0x2a:
We found that the slack in goth does not get the userinfo correctly : https://github.com/casdoor/casdoor/issues/618 . After consulting the two api documents ([auth.test](https://api.slack.com/methods/auth.test), [users.info](https://api.slack.com/methods/users.info)), we found that currently the slack does not allow passing a token in the get parameter and has to use the body of the post to pass it or use the bearer token.

This pr modifies the slack request method and will now use the bearer token for the request.


Steve0x2a:
@markbates @techknowlogick  plz review

PR #456, opened 2022-05-06
fix(lastfm): fix auth callback param

jackmerrill:
- Fixes callback url by changing "callback" to "cb"
- Uses ".com" English website


techknowlogick:
Thanks for the PR Jack :)

Could you update the tests as well?

jackmerrill:
> Thanks for the PR Jack :)
> 
> Could you update the tests as well?

I just did! Sorry 'bout that haha.

PR #459, opened 2022-06-06
zoom: allow overriding redirect_uri and passing code_verifier 

mxaly:
This PR extends zoom provider Session.Authorize by making it possible to pass `redirect_uri` (that will later on override redirect_uri from config) and `code_verifier` params. 

`verifier_code` param is required in order to support PKCE in zoom 
`redirect_uri` override is required as in some cases (for eg when using zoom in-app-auth) zoom is starting the authorization process and sets the redirect_uri dynamically.

### DIFF
--- a/providers/zoom/session.go
+++ b/providers/zoom/session.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/markbates/goth"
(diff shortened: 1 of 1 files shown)


mxaly:
Thanks @techknowlogick :)

PR #461, opened 2022-06-23
feat: add user email in kakao

minhenry747:
Added Kakao's user email when fetchuser.


techknowlogick:
Thanks for this PR, will the struct field be nil if the scope is not passed?

PR #481, opened 2022-11-08
Fix Patreon FetchUser

Jleagle:
Without this the fetch user call does not return all the info to fill in User{}


Jleagle:
Hey @techknowlogick, could i get this merged in please, currently these fields do not get set (created, email, full_name, image_url, vanity). Thanks!

techknowlogick:
@Jleagle thanks for ping :)

Issue #483, opened 2022-11-23
Mastodon OAuth implementation has no handling for the email field.


chizutan5:
According to my code reading skills, the mastodon oauth implementation on Goth has no handing for the email field, this creates what I believe is a bug on Gitea where you cannot create a gitea account with a mastodon oauth session because it will ask for the email, however you can link a mastodon authententication session to an existing gitea account which already will have an email attached.
I have documented this issue at go-gitea/gitea#21710

Pleroma (or akkoma/newroma/rebased) instances and possibly mastodon too (I am not able to test), which are otherwise compatible with mastodon auth clients, will not provide an email to the client and as such cause this problem.
We should take the user's fediverse userid (user@instance.tld) and apply this to the email field, in the same way as pronouns.page handles its own custom mastodon login, or optionally allow the user to enter their own email address in.

I would offer to pull request but I don't know the first thing about go, I was only barely able to get a custom php implementation with mediawiki working with much outside help.

PR #490, opened 2022-12-21
Attempted fix for #483

chizutan5:
I have attempted to use the strategies from the discord oauth (which has email handing) for mastodon, to provide the oauth client a usable email in the form of the user fqn (user@instance.tld) which should allow gitea to create user accounts through oauth, please correct if needed, but I think this might work.

### DIFF
--- a/providers/mastodon/mastodon.go
+++ b/providers/mastodon/mastodon.go
@@ -153,6 +153,7 @@ func userFromReader(r io.Reader, user *goth.User) error {
 		NickName  string `json:"username"`
 		ID        string `json:"id"`
 		AvatarURL string `json:"avatar"`
+		Email     string `json:"fqn"`
 	}{}
 	err := json.NewDecoder(r).Decode(&u)
 	if err != nil {
@@ -165,6 +166,7 @@ func userFromReader(r io.Reader, user *goth.User) error {
 	user.NickName = u.NickName
 	user.UserID = u.ID
 	user.AvatarURL = u.AvatarURL
+	user.Email = u.Email
 	return nil
 }
 


chizutan5:
@techknowlogick Are you able to take a look at this and advise if something is wrong?

techknowlogick:
Thanks for this PR @chizutan5, I'm unsure about merging this PR as some systems expect that when being passed something via the email field that it is indeed the users field. This may lock out users from their account as they wouldn't have access to an email that doesn't exist. Is there a different way to get the email from mastodon, perhaps by passing an additional scope?

chizutan5:
> Thanks for this PR @chizutan5, I'm unsure about merging this PR as some systems expect that when being passed something via the email field that it is indeed the users field. This may lock out users from their account as they wouldn't have access to an email that doesn't exist. Is there a different way to get the email from mastodon, perhaps by passing an additional scope?

Hi,
Mastodon, Pleroma and possibly Misskey don't actually provide emails at all from what I can find, and there isn't any alternative scopes I know of that support giving it out, this is the only thing that provides an email-like ID which can be used.
Maybe it can be passed on, specifically as "FQN", but this isn't compatible and for any software that needs an email but hasn't had any changes to support this fqn tag.
I would argue that oauth generally negates the need to worry about users being locked out, as if they are, then they usually need to deal with the oauth provider who does have some kind of external contact method.

Maybe it can be possible to send some additional data that can then be interperated by compatible implementations, that tell them "hey, this is an activitypub id" so that it can be handled in some way but still allow legacy implementations to take it into the email field? This was originally a PR to fix a gitea issue where it wasn't possible to create an account through the mastodon oauth, is it better fixed over there by creating an FQN field that satisfies the requirements for needing an email on the account? (Then this PR can just be changed to feed through the FQN instead of claim that it is an email)

It's definitely off topic discussion but I have thought about how gitea could act as an activitypub server and send out activitypub direct messages to activitypub ids instead of emails so that that sort of fix actually works as it might be expected, maybe this oauth implementation is better suited for directly implementing into gitea instead of changing anything here?

I appreciate your feedback, since this could be a breaking change for some software, but I don't have context for what actually uses goth other than gitea.

Thank you.

PR #491, opened 2022-12-22
Add cognito provider

pilotso11:
Added support for AWS Cognito.
New takes 3 parameters all from the Cognito console:
- The client ID
- The client secret
- The base URL for your servcice, either a custom domain or cognito pool based URL
You need to ensure that the source login URL is whitelisted as a login page in the client configuration in the cognito console.
GOTH does not provide a full token logout, to do that you need to do it in your code, my recommendation is
a) Destroy your session (or however else you handle the logout internally)
b) redirect to https://customdomain.auth.us-east-1.amazoncognito.com/logout?client_id=clinet_id&logout_uri=http://localhost:8080/    (or whatever your login/start page is).
c) Note that this page needs to be whitelabeled as a logout page in the cognito console as well.

This has been tested using goth_fiber.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,7 @@ on:
   pull_request:
     branches:
     - master
+  workflow_dispatch:
 
 name: ci
 
--- a/providers/cognito/cognito.go
+++ b/providers/cognito/cognito.go
@@ -13,6 +13,19 @@ import (
 )
 
 // Provider is the implementation of `goth.Provider` for accessing AWS Cognito.
+// New takes 3 parameters all from the Cognito console:
+// - The client ID
+// - The client secret
+// - The base URL for your servcice, either a custom domain or cognito pool based URL
+// You need to ensure that the source login URL is whitelisted as a login page in the client configuration in the cognito console.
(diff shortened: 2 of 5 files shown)


pilotso11:
I removed the url.JoinPath which broke the CI - it should pass tests on pre 1.19 version of go now.

pilotso11:
Ok - reverted from io.ReadAll to ioutil.ReadAll   (fingers crossed this should build on 1.15).
I also add a manual trigger to the ci.yml, this would let me riun the CI on my dev branch in my fork.  
I'm struggling to reproduce the CI build error locally - even though its legit as the function moved in 1.16 so it should fail on 1.15, so the ability to run the CI before you kick it off would be handy.

pilotso11:
You might want to rerun the CI.  It failed with what looks very odd.  Maybe some random env issue.  All the tests in the providers actually passed.  I tried a pull request on my fork to kick off the CI and it all worked that for all the builds on that run:  https://github.com/pilotso11/goth/pull/1

techknowlogick:
Thanks so much @pilotso11 <3 and for working through the issues with CI :)

PR #492, opened 2022-12-24
Update Instagram scope

mrpkin:


providers/instagram/instagram.go
```diff
@@ -19,7 +19,7 @@ import (
 var (
 	authURL         = "https://api.instagram.com/oauth/authorize/"
 	tokenURL        = "https://api.instagram.com/oauth/access_token"
-	endPointProfile = "https://api.instagram.com/v1/users/self/"
+	endPointProfile = "https://graph.instagram.com/v15.0/me"
```
techknowlogick, line 22: Is it still instagram.com for the API endpoints? From docs it looks like these have been updated to `facebook.com`

I'm looking at https://developers.facebook.com/docs/instagram-api/getting-started

PR #499, opened 2023-02-26
fixed error of bitbucket in struct

nik0811:
fixed error of bitbucket while parsing the callback response
Not able to parse email properly so login was failling now fixed

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/nik0811/goth
+module github.com/markbates/goth
 
 go 1.15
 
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/markbates/goth
+module github.com/nik0811/goth
```
techknowlogick, line 1: Thanks for PR, are you able to revert this change?

nik0811: @techknowlogick  yes reverted,

techknowlogick:
Thanks again :) I've just merged these changes.

PR #504, opened 2023-03-30
Allow for multiple OIDC providers

yaronius:
OpenID Connect is pretty generic so we might want to use several different providers at the same time. But it's not possible at the moment because the provider is mapped to a single name. It would make sense to make this configurable. What do you think @techknowlogick?


techknowlogick:
Thanks for the PR @yaronius, I made a minor update to use a switch case instead of an if conditional

PR #518, opened 2023-08-02
Fix empty raw data in steam provider

leobrines:
Fixed empty raw data in user response when you use Steam provider.

### DIFF
--- a/providers/steam/steam.go
+++ b/providers/steam/steam.go
@@ -185,6 +185,17 @@ func buildUserObject(r io.Reader, u goth.User) (goth.User, error) {
 		u.Location = "No location is provided by the Steam API"
 	}
 
+	// Set raw data
+	b, err := json.Marshal(&player)
+	if err != nil {
+		return u, err
+	}
+
+	err = json.Unmarshal(b, &u.RawData)
+	if err != nil {
+		return u, err
+	}
+
 	return u, nil
 }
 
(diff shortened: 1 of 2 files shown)


yabanci:
Hi @leobrines — hope you don't mind, I picked this up to try to land it. The gap you identified is still present on master 2+ years later, so the underlying fix is still needed.

The follow-up PR is #644. Two changes vs. your original patch, with reasoning:

1. **Decode the body once and unmarshal the player into both the typed struct and `u.RawData`** instead of marshalling the typed struct back to JSON and decoding that into the map. Your approach worked correctly but it made `RawData` a strict duplicate of the six typed fields (steamid, personaname, realname, avatarfull, loccountrycode, locstatecode) — so callers still couldn't reach `communityvisibilitystate`, `timecreated`, `primaryclanid`, `profileurl`, etc. The new approach exposes the full Steam player payload, which is what `RawData` does in every other goth provider (auth0, discord, etc.).

2. **Expanded the test fixture** beyond the six fields you used to a realistic Steam player payload, and added `RawData` assertions for the no-slot fields. This way a future refactor can't silently shrink the response again — which is, I suspect, exactly the kind of latent regression that left this PR sitting for two years.

Your `httpTestTransport` indirection and the `testUserSummaryBody` structure are preserved as-is, those were already the right design. Credit to you for the original analysis and the test infrastructure — both still load-bearing here.

Suggest closing this in favour of #644 once it lands.

leobrines:
closed in favor of @yabanci changes in https://github.com/markbates/goth/pull/644

PR #526, opened 2023-10-27
chore: slice loop replace

testwill:


techknowlogick:
Thanks for the PR. GPlus is deprecated (and now removed) so I will close this PR.

PR #531, opened 2023-12-09
Adding a Hubspot provider for Goth

justinrsweeney:
This adds Hubspot as a Goth provider: https://developers.hubspot.com/docs/api/working-with-oauth.

### DIFF
--- a/providers/hubspot/hubspot.go
+++ b/providers/hubspot/hubspot.go
@@ -1,0 +1,175 @@
+package hubspot
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/markbates/goth"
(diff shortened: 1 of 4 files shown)


justinrsweeney:
@techknowlogick Looks like the test failure here is related to building with Go 1.16 and upgrading to 1.17 would resolve the issue. It seems to be unrelated to the specific changes here, curious if you have a suggestion on how to proceed?

techknowlogick:
Thanks @justinrsweeney, I've just bumped the min version to 1.18. could you merge in the recent changes to your branch so it can run again?

justinrsweeney:
Thanks, updated from the recent changes.

PR #538, opened 2024-02-07
Fix occasional error while unmarshalling Apple id_token

yaronius:
Hi! According to [Apple documentation on the ID token](https://developer.apple.com/documentation/sign_in_with_apple/sign_in_with_apple_rest_api/authenticating_users_with_sign_in_with_apple#3383773) it turns out that some claims could be either boolean or string representing boolean. In this library we always parse one of these claims (`is_private_email`) as a string (notice the tags on [this line](https://github.com/markbates/goth/blob/master/providers/apple/session.go#L54)) and sometimes (when Apple actually sends boolean value) it results in the following error: `json: invalid use of ,string struct tag, trying to unmarshal unquoted value into bool`. I recently noticed an increase in the amount of such errors, probably now Apple uses boolean more frequently. This PR fixes that by detecting the value from token and handling it appropriately.


yaronius:
Not sure why tests are failing 🤔

techknowlogick:
thanks @yaronius, could you merge in the recent push to master, as I've updated stuff around tests/min go verson

techknowlogick:
Thanks!

techknowlogick:
I'll (hopefully) cut a release soon.

dstapleton92:
This has been driving us crazy! Thanks so much for the fix. Also looking forward to a release soon!

techknowlogick:
@dstapleton92 thanks for the reminder. I just cut the release.

PR #539, opened 2024-02-10
add providers resolver

titusgahissy:
Providers config (`client_id`, `client_secret`, etc) can be stored in a database and change while the application is running.
The current syntax does not allow to handle this usecase.

```go
goth.UseProviders(providers) // static data
```

This PR introduces a new interface:

```go
type ProviderResolver interface {
  Get(name string) (Provider, error)
  GetAll() Providers
}
```

```go
// For backward compatibility, goth.UseProviders(providers) now becomes :

func UseProviders(viders ...Provider) {
	providers := Providers{}
	for _, p := range viders {
		providers[p.Name()] = p
	}
	resolver = DefaultProviderResolver{providers: providers}
}
```

Now it becomes possible to do something like:

```go

type MyCustomResolver struct{
  db *gorm.DB
}

func (r MyCustomResolver) Get(name string) (goth.Provider, error) {
   // do your magic
}

func (r MyCustomResolver) GetAll() goth.Providers {
  // do your magic
}

goth.SetProviderResolver(&MyCustomResolver{db: db})
```

### DIFF
--- a/provider.go
+++ b/provider.go
@@ -26,36 +26,66 @@ const NoAuthUrlErrorMessage = "an AuthURL has not been set"
 // Providers is list of known/available providers.
 type Providers map[string]Provider
 
-var providers = Providers{}
-
-// UseProviders adds a list of available providers for use with Goth.
-// Can be called multiple times. If you pass the same provider more
-// than once, the last will be used.
-func UseProviders(viders ...Provider) {
-	for _, provider := range viders {
-		providers[provider.Name()] = provider
-	}
+// ProviderResolver an interface to resolve a provider by name
+type ProviderResolver interface {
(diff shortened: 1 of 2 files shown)


techknowlogick:
Thanks @titusgahissy, could you resolve the conflicts?

go.mod
```diff
@@ -1,41 +1,20 @@
 module github.com/markbates/goth
```
techknowlogick, line 1: thanks for updating this PR. It looks like somethings have been reverted from master in go.mod. Could you bump the version to 1.18, and run `go mod tidy`?

titusgahissy: should be ok now.

PR #540, opened 2024-02-13
FIX apple sign in

hjanuschka:
apple turned this into a real bool, and unmarshaling fails using the string thing


yaronius:
This will fix some cases but will break others. I suggested a fix that covers all cases from Apple documentation https://github.com/markbates/goth/pull/538

hjanuschka:
awesome ❤️ - closing in favor of #538

PR #546, opened 2024-03-24
gothic: fix link to examples

adam000:
I noticed that these links were broken when using the documentation, so I thought I'd fix it up really quickly :)


techknowlogick (approved):
Thanks!

PR #523, opened 2023-10-01
Add reddit provider and session

ccaneke:
Information on reddit oauth2 can be found [here](https://github.com/reddit-archive/reddit/wiki/OAuth2).

PR #568, opened 2024-07-09
Added Reddit to the Supported Providers list

### LINKED ISSUE #532: reddit provider not found
I see the reddit provider is in the providers directory, but it doesn't seem to be exported as a package for use. Is there some problem with it, or is that just an oversight?

Celarye:
Accompanies #523 and closes #532


techknowlogick (approved):

techknowlogick:
Thanks!!!!🙏

PR #569, opened 2024-07-13
Adding get provider name support for chi router

fardinabir:
[Chi](https://github.com/go-chi/chi), one of the fastest and most lightweight routers for Go. When it is used for routing requests to a Gothic handler like below, Gothic fails to get the value of the 'provider' parameter.
```go
func signUpRouter() *chi.Mux {
	r := chi.NewRouter()
	r.Get("/{provider}", gothic.BeginAuthHandler)
	return r
}
```
The way to get the parameter value from a Chi request is `chi.URLParam(r, paramName)`. However, inside the `getProviderName` function of Gothic, there is no support for this extraction method.

By merging this PR, it will no longer be necessary to set the provider explicitly in the context of the request to get Gothic working with the Chi router.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/markbates/goth
 go 1.18
 
 require (
+	github.com/go-chi/chi/v5 v5.1.0
 	github.com/golang-jwt/jwt/v4 v4.2.0
 	github.com/gorilla/mux v1.6.2
 	github.com/gorilla/pat v0.0.0-20180118222023-199c85a7f6d1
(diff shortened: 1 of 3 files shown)


techknowlogick:
Thanks!!! I use chi myself and have just been using a work around until now. This'll hopefully let me refactor that code away :)

lrstanley:
I don't understand why this was merged. Is there a reason goth wants to require everyone pull in chi even if they don't use it? There are no other routing libraries pulled in, so why chi? This doesn't make sense to me.

PR #570, opened 2024-07-26
Add 1.22 mux support

dgduncan:
This PR seems large; however, it fundamentally accomplishes 3 things.

1. Upgrade GO version to `1.22`
2. Add compatibility for the new 1.22 mux functionality to retrieve provider values
3. Removes all instances of `ioutil` with their non-deprecated `io` and `os` alternatives.

This this change and without any other modifications, this change breaks backward compatibility for GO <1.22. This can be fixed however by creating a new `gothic.go | gothic_test.go` with `//go:build go1.22` with these breaking changes. I am happy to make these compatibility changes but would love to get other's opinions first.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [ 1.22.x ]
+        go-version: [ 1.18.x, 1.19.x, 1.20.x, 1.21.x, 1.22.x ]
         os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [ 1.18.x, 1.19.x, 1.20.x, 1.21.x, 1.22.x ]
+        go-version: [ 1.22.x ]
         os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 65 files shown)


techknowlogick (approved):
Thanks for this PR!

One tiny nitpick: I'm hesitant to bring the minimum up to 1.22 as goth is used by many downstreams, some of which still have much lower minimums. Would you be able to add in a polyfill for the new capabilities if the library is being used with a version of go that doesn't support them?

Meta: I do recognize the weirdness of asking for this as it means folks using older versions of go would be upgrading goth but not go (I've long given up on trying to understand some upgrade policies, and if an accommodation can be made without impact to security or significantly increasing maintenance burden, then I do try to at least investigate)

dgduncan:
@techknowlogick I am open to ideas. I extracted out the `getProvider` functionality into two new files; each tagged with a version check. I would love to see if this now passes the tests of the pipeline. If you have a better idea for how to implement such backwards compatibility I would love to know. I am not super familiar with this type of backwards compatability.

dgduncan:
Ok it looks like my change passes the tests; however, I still want to do some individual testing. Also I do not know how I feel about the `provider_legacy` naming. @techknowlogick do you have an opinion?

majst01:
maybe splitting into two PRs, one with the removal of ioutl which deprecated since 1.16, and one with the mux introduction which only works >= 1.22.

dgduncan:
@techknowlogick I have noticed that this PR has been sitting here a while. To make it more palatable of a merge, would you like me to do as @majst01 mentioned and split this into two smaller PRs. I understand that this is kind of a beefy PR that does has some potential risk with the changes asked.

dgduncan:
I have created a companion PR that splits out the `io/ioutil` changes in order to potentially decrease the blast radius of this PR.

https://github.com/markbates/goth/pull/600

PR #573, opened 2024-08-05
adding get provider name support for go 1.22 servemux

mowazzem:


techknowlogick:
Thanks for your PR! This is currently in progress in https://github.com/markbates/goth/pull/570 and should be merged shortly.

PR #98, opened 2016-09-05
Add Intercom provider

fox:
This adds support for [Intercom](https://intercom.io), with tests and example.

goth-intercom

PR #583, opened 2024-11-23
feat: upgrade golang-jwt version for apple provider

burkayanduv:
I was getting `token used before issued` error with `golang-jwt` `v4`, so I've upgraded it to `v5`. Here is the reference to the issue in the source repository:
[Token used before issued #98](https://github.com/golang-jwt/jwt/issues/98)

This PR updates the Apple provider integration by upgrading the golang-jwt library. Key changes include:

- Upgraded golang-jwt from `v4.2.0` to `v5.2.1`.
- Replaced deprecated methods with the new `jwt.NewValidator` API for token validation.
- Updated returned error type as `jwt.ValidationError` was removed in the latest version.

The change in the returned errors will break backwards compatibility if someone has a check on returned error, but that happened in the golang-jwt as well.


techknowlogick:
Thanks!!

PR #585, opened 2024-12-06
Added UseFilesystem and UseCookies for sessions.Store

andreimerlescu:


andreimerlescu:
I need this for a package I am working on where the upstream goth is not working, but I will continue to use this master branch for my project. Others can use it if they need; this patch includes adding sessions.FilesystemStore and it introduces 2 new dependencies that are pure Go packages called https://github.com/andreimerlescu/go-checkfs and https://github.com/andreimerlescu/go-passwd ; the former validates the directory has the proper permissions on it ; but is wrapped in a very clean way and is very inexpensive but highly expandable; I am using it throughout my work and I found that packages like this are difficult to work with when they are not passing errors through properly; so I am going to continue refactoring this project and it will end up being quite a deviation from the upstream. If the team is interested in this fork, they can reach out.

andreimerlescu:
Ugh I didn't want to push it here. Github is annoying right now. User error. My bad. I need to work off a fork until I very its working and then I'll push. Closing this now.

Issue #616, opened 2025-08-18
Apple auth not fetching first name and last name even with the name scope


Erol:
It seems like the first name and last name is not being fetched for Apple Sign-In even if the `name` scope has been specified. I tried looking at the code it mentioned: https://github.com/markbates/goth/blob/4b34e1719daae56def96b42ec23ca05badf6ca24/providers/apple/apple.go#L124-L130

Since `CompleteUserAuth` has access to the request, I was thinking that fetching the names should be possible, though there would be a change on the method signature of `FetchUser`.

Just wondering if the omission was more of a limitation or was this intentional? If it's a limitation with the current implementation I'd like to take a stab at implementing this.

PR #624, opened 2025-08-22
Get user full name from Apple Sign-In

Erol:
The user full name should be available to us as form parameters from the request when we make the call to `sess.Authorize`. This implements getting the user full name if it is available and should resolve https://github.com/markbates/goth/issues/616


techknowlogick:
@Erol It had to be manually done. I think there might be some threshold for it to be automatically approved (not sure about github's rules on this as some repos don't have this). Regardless, I've approved them for being run (I'm unsure if you push again if it needs to be manually approved again though)

PR #637, opened 2026-03-15
Add RP-Initiated Logout support for OpenID Connect provider

tphakala:
## Summary

- Add optional `LogoutProvider` interface to goth with `EndSessionURL` method
- Implement it on the `openidConnect` provider to build `end_session_endpoint` URLs per the [OpenID Connect RP-Initiated Logout 1.0 spec](https://openid.net/specs/openid-connect-rpinitiated-1_0.html)
- The `EndSessionEndpoint` was already discovered from `.well-known/openid-configuration` and stored in `OpenIDConfig` but was never exposed — this makes it usable

### Design

The `LogoutProvider` interface is optional and doesn't change the existing `Provider` interface, so this is fully backward compatible. Consumers check support via type assertion:

```go
if lp, ok := provider.(goth.LogoutProvider); ok {
    logoutURL, err := lp.EndSessionURL(user.IDToken, "http://localhost/post-logout", state)
    // redirect user to logoutURL
}
```

Parameters follow the spec:
- `id_token_hint` — the ID token issued to the user (recommended)
- `post_logout_redirect_uri` — where to redirect after logout (optional)
- `state` — CSRF protection (optional)
- When no `id_token_hint` is provided, `client_id` is included as fallback per spec

## Test plan

- [x] Tests for `EndSessionURL` with all parameter combinations
- [x] Test for missing `end_session_endpoint` returns clear error
- [x] Test that provider implements `LogoutProvider` interface
- [x] All existing tests pass
- [x] `golangci-lint` clean


tphakala:
Opened against wrong repo, closing.

PR #640, opened 2026-04-23
feat(openidConnect): auto-detect and apply PKCE from OIDC discovery

### LINKED ISSUE #639: openidConnect: auto-configure PKCE based on the contents of the discovery document
I'm having issues setting up a custom oidc auth provider in Gitea using the goth openidConnect provider. The auth requests are denied because my provider mandates the use of PKCE.

This could be configured at the provider level according to #473 I think, but ideally the feature should be picked up from the discovery document: the presence of a non-empty `code_challenge_methods_supported` in the json response signals that PKCE is available, and lists the supported challenge methods. If the provider and the RP are spec-compliant, it should "just work" without additional config.

peyremorgan:
When the OIDC discovery document advertises code_challenge_methods_supported, the provider now automatically:
- Selects the best method (S256 preferred over plain per RFC 7636 §4.2)
- Generates a cryptographically random code_verifier on each BeginAuth call
- Computes the code_challenge and injects it into the authorization URL
- Stores the code_verifier in the Session for use at token exchange
- Uses the stored verifier in Authorize(), falling back to the legacy code_verifier query param for backward compatibility

If code_challenge_methods_supported is absent or empty, PKCE is not applied.

Fixes #639

### DIFF
--- a/providers/openidConnect/openidConnect.go
+++ b/providers/openidConnect/openidConnect.go
@@ -2,6 +2,8 @@ package openidConnect
 
 import (
 	"bytes"
+	"crypto/rand"
+	"crypto/sha256"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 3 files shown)


peyremorgan:
Okay, marking this as WIP because #636 uses another approach (imports "golang.org/x/oauth2") and I think we should converge on a single implementation style for both.

Issue #430, opened 2021-10-21
Auth0 provider doesn't attribute all the values to goth.User


justman00:
Hey,

so I noticed that Auth0 is sending some more data back, e.g:

```json
{
        "email": "email@example.com",
        "email_verified": true,
        "family_name": "Example",
        "given_name": "Example",
        "locale": "en",
        "name": "Example Example",
        "nickname": "example.example",
        "picture": "<picture url>",
        "sub": "google id or other provider id",
        "updated_at": "2021-10-21T08:52:15.784Z"
    }

```

but the provider seems to be reading only [these values](https://github.com/markbates/goth/blob/master/providers/auth0/auth0.go#L35):
```golang
type auth0UserResp struct  {
	Name      string `json:"name"`
	NickName  string `json:"nickname"`
	Email     string `json:"email"`
	UserID    string `json:"sub"`
	AvatarURL string `json:"picture"`
}
```

I want to know if other values, such as **family_name** or **given_name** are not parsed on purpose because they can be left empty by **Auth0** or if it was just overlooked?

yabanci:
@justman00 To answer your question directly: it was an oversight, not intentional.

`given_name` and `family_name` are part of the OIDC standard claim set and Auth0 returns them whenever the user provided a name. The Google provider already maps them to `goth.User.FirstName` / `goth.User.LastName` (`providers/google/google.go`) — Auth0 should do the same. Picture, name, email, sub, and nickname were mapped; given_name and family_name just got dropped.

Opened #643 with the fix. It adds both fields to `auth0UserResp` and `userFromReader`. The unmapped claims you also listed (`email_verified`, `locale`, `updated_at`) don't have typed slots on `goth.User`, so they stay accessible via `user.RawData` — added an assertion to `Test_FetchUser` to lock that contract in.

pooya2831:
A

Issue #414, opened 2021-07-29
discord provider requires "identify" scope


tardisx:
This might be a fairly naive problem on my behalf, this is fairly new to me.

I have successfully used the google provider with just the "email" scope.

When I try the same with discord, when the callback is called I get an error:

    discord responded with a 401 trying to fetch user information

I suspect the returned body of the request made in FetchUser has something interesting, but I'm not familiar enough with net/http yet to work out how to inject my own client which can expose that.

If I add the "identify" scope, things work correctly.

yabanci:
@tardisx Not naive at all — this is a real bug, just well-hidden by Discord's docs.

The reason is in `providers/discord/discord.go:newConfig`. When you pass any custom scope (e.g. just `email`), the code throws away `ScopeIdentify` and only uses what you passed. But Discord's `/users/@me` endpoint requires `identify` for access at all — `email` is supplementary (it enriches the response, doesn't grant access). Without `identify`, you get the 401 you saw.

Confirmed with Discord's current docs: *`identify` allows /users/@me without email* and *`email` enables /users/@me to return an email* — i.e. `email` presupposes `identify`.

Fix in #642: `newConfig` now always starts with `ScopeIdentify` and dedupes when the caller passes it explicitly. Added a table-driven regression test covering email-only, explicit identify, and multi-scope combinations.

PR #641, opened 2026-05-17
fix(vk): tolerate accounts without email (#338)

### LINKED ISSUE #338: VK provider returns error for users with empty 'email' field
Hi
I've encountered an issue where `vk` provider will return "Cannot fetch user email" error for users which doesn't have this field filled in their VK profile.
I can't see why this is treated as an error and I think it should be fixed.

I can do a pull request fixing the issue but not sure what is the best way to implement it.
Fixing it directly in `vk` provider will probably break compatibility for some of existing goth users. I think some method on `vk` provider or separate provider should be created. Any opinions?

Steps to reproduce the issue:
1) register new VK account without specifying email. (I think it became possible since around 2018, but may be earlier).
2) Try to authorize him with `goth/provider/vk`

Expected:
`gothic.CompleteUserAuth(c.Writer, c.Request)` should return no error.

Actual:
`CompleteUserAuth` returns error `Cannot fetch user email`

yabanci:
## What

Closes #338.

VK's OAuth2 token endpoint omits the top-level `email` field when the user has no email in their VK profile (VK has allowed registration without email since ~2018). The current code treats this as a fatal error:

```go
email, ok := token.Extra("email").(string)
if !ok {
    return "", errors.New("Cannot fetch user email")
}
```

…which means accounts without an email can't complete OAuth at all.

## Change

Make `email` optional in `providers/vk/session.go`:

```go
if email, ok := token.Extra("email").(string); ok {
    s.email = email
}
```

- Accounts **with** email: unchanged — `Session.email` is populated and propagated to `goth.User.Email`.
- Accounts **without** email: no longer fail; `goth.User.Email` is empty, matching the contract already used by providers like Twitter where email is opt-in.

## Backward compatibility

The original issue author (@nikita-vanyasin) wondered whether fixing this in-place would break existing users. In practice it cannot: the only callers affected are those that were already getting `"Cannot fetch user email"` and thus already broken. Callers that received an email continue to receive it.

If a downstream caller really needs to *enforce* an email, they can already do so by checking `user.Email == ""` after `gothic.CompleteUserAuth` — which is the standard pattern for optional-email providers.

## Tests

Added `providers/vk/authorize_internal_test.go` with two subtests using `httptest.Server` to mock the token endpoint:

- `TestAuthorize_PopulatesEmailWhenPresent` — happy path, regression-guards existing behaviour.
- `TestAuthorize_NoErrorWhenEmailMissing` — fails on master, passes with this fix.

Verified locally: the new regression test fails when the patch is reverted and passes with it applied. Full `go test ./...` is green; `go vet ./...` is clean.

### DIFF
--- a/providers/vk/authorize_internal_test.go
+++ b/providers/vk/authorize_internal_test.go
@@ -1,0 +1,68 @@
+package vk
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// tokenServer spins up an httptest.Server that responds to the OAuth2 token
+// exchange with the given JSON body, and swaps the package-level tokenURL so
+// the provider hits it. The returned cleanup restores the original tokenURL
+// and shuts the server down.
+func tokenServer(t *testing.T, body string) func() {
+	t.Helper()
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		fmt.Fprint(w, body)
+	}))
+	original := tokenURL
+	tokenURL = server.URL
+	return func() {
+		tokenURL = original
+		server.Close()
+	}
+}
+
+func TestAuthorize_PopulatesEmailWhenPresent(t *testing.T) {
+	t.Parallel()
+	a := assert.New(t)
+
+	cleanup := tokenServer(t, `{"access_token":"tok","expires_in":86400,"user_id":1,"email":"user@example.com"}`)
+	defer cleanup()
+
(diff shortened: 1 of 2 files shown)


yabanci:
Hi @techknowlogick — gentle nudge on a small batch of bug-fix PRs I opened in May, in case they're useful for your next cleanup pass. All four are surgical, close an open issue, and are conflict-free against current master:

- #641 — `fix(vk)`: tolerate accounts without email (closes #338)
- #642 — `fix(discord)`: always include identify scope (closes #414)
- #643 — `fix(auth0)`: map given_name/family_name to User (closes #430)
- #644 — `fix(steam)`: populate RawData from player payload (revives #518, closes #518)

No rush — happy to rebase or split any of them if that makes review easier. Thanks for maintaining goth!

